PR #4, opened 2019-10-12
EHR-534 prototype for AQL json result set comparison

ppazos:


Geisterfalle:
Would be good if we can briefly go over these changes together

PR #7, opened 2019-10-15
Canonical JSON returned for partial aql path (NB. full composition is…

chevalleyc:
… not yet supported)

### DIFF
--- a/service/src/main/java/org/ehrbase/aql/sql/binding/SelectBinder.java
+++ b/service/src/main/java/org/ehrbase/aql/sql/binding/SelectBinder.java
@@ -147,25 +147,25 @@ public class SelectBinder extends TemplateMetaData implements I_SelectBinder {
     }
 
 
-    public PathResolver getPathResolver() {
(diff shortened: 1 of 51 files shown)


service/src/main/java/org/ehrbase/aql/sql/binding/SelectBinder.java
```diff
@@ -147,25 +147,25 @@ public Condition getWhereConditions(String templateId, UUID comp_id) {
...
+//        return compositionAttributeQuery.getEhrIdAlias();
+//    }
+//
+//    public boolean isCompositionIdFiltered() {
+//        return compositionAttributeQuery.isCompositionIdFiltered();
+//    }
+//
+//    public boolean isEhrIdFiltered() {
+//        return compositionAttributeQuery.isEhrIdFiltered();
+//    }
```
stefanspiska: If not needed anymore it should be removed.

Geisterfalle:
It now part #31

PR #17, opened 2019-10-23
adds more unit-tests for jsonlib

testautomation:
this adds some exception handling and a few more unit-tests for jsonlib.py


testautomation:
image

PR #19, opened 2019-10-24
7 run and check all tests in folder access test.java against the real db

lmarcoruiz:
corrects some bugs and cascade deletions that were present when testing against the real DB. Also standardizes line changes so the FolderAccessTests run in all platforms.
@Reviewer please confirm the latter works in Windows before approval.


lmarcoruiz:
insertion and update test still fail due to OS differences in blanks and line changes. They will need to be corrected in a windows machine.

PR #25, opened 2019-11-01
fixes XML validation issue in tests - ehrbase/project_management#33

testautomation:
solves issue ehrbase/project_management#33


ppazos:
looks good!

PR #29, opened 2019-11-07
Feature/ehr 452 update folders service

axelsiebert:


service/src/main/java/org/ehrbase/dao/access/jooq/FolderAccess.java
```diff
@@ -637,6 +637,41 @@ public static I_FolderAccess buildFolderAccessForInsert(
         return folderAccess;
     }
 
+    public static I_FolderAccess buildUpdateSubFolderAccess(
```
lmarcoruiz: why do we need this method, aren´t we replicating functionality here?

PR #31, opened 2019-11-08
Feature/ehr 163 support canonical for datavalue

chevalleyc:
This is a super set of EHR-450 with fixes to integrate the requested change, plus fixed EHR-163, EHR-159, EHR-160, EHR-145, full composition support.
NB. There are still some issues with AQL expression based on full composition content (f.e. a/content[xyz]) however this is circumvented by querying a known item (f.e. observation, section etc. defined by a contains clause)

### DIFF
--- a/base/src/main/resources/db/migration/V11__raw_json_encoding_new_format.sql
+++ b/base/src/main/resources/db/migration/V11__raw_json_encoding_new_format.sql
@@ -1,0 +1,418 @@
+/*
+ * Modifications copyright (C) 2019 Vitasystems GmbH and Hannover Medical School.
+
+ * This file is part of Project EHRbase
+
+ * Copyright (c) 2015 Christian Chevalley
+ * This file is part of Project Ethercis
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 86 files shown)


chevalleyc:
> When creating the following query:
> 
> `{ "q": "select a as test from EHR e contains COMPOSITION a contains OBSERVATION a_a[openEHR-EHR-OBSERVATION.time_series.v1]" }`
> 
> I find the following format in the path and its name:
> 
> `{ "q": "select\n a as test\nfrom EHR e\ncontains COMPOSITION a\ncontains OBSERVATION a_a[openEHR-EHR-OBSERVATION.time_series.v1]", "name": null, "columns": [ { "path": "/ehr.js_composition(composition_join.id,'local.ethercis.com')::text", "name": "#0" }`
> 
Please check with the updates, name and path should be consistent with the specs.
> The following query:
> 
> `{ "q": "select a_a/data[at0001]/events[at0002]/data[at0003]/item from EHR e contains COMPOSITION a contains OBSERVATION a_a[openEHR-EHR-OBSERVATION.time_series.v1]" }`
> 
> provides the following result:
> 
> `{ "q": "select\n a_a/data[at0001]/events[at0002]/data[at0003]/item\nfrom EHR e\ncontains COMPOSITION a\ncontains OBSERVATION a_a[openEHR-EHR-OBSERVATION.time_series.v1]", "name": null, "columns": [ { "path": "/FIELD_0", "name": "#0" } ], "rows": [ [ null ], [ null ], [ null ] ] }`
> 
> I think this should provide a single row with null.
The issue is that the provided path `a_a/data[at0001]/events[at0002]/data[at0003]/item` is not correct and depending on your data this is expanded into an array search (guess with a valid path, 3 values/structure should be return). A *clean* fix would be to actually validate the path before executing the query. I suggest to open a ticket for this.

PR #36, opened 2019-11-12
Bugfix for #43 PGObject cannot be cast to ItemStructure

axelsiebert:
- Extended Jakes OtherDetailsJsonbBinder to also bew aware of details column for folders
- Replaced usage of PGObectParser by new binded ItemStructure
- Updated Tests for usage of ItemStructure.

@jakesmolka: 
Please give a short feedback if I correctly understand the usage of the binder. I added the regular expression as discussed. I also made them more strict since the regex `.*details.*` would break the generation of "audit_details". If we want to add more ItemStructure columns here, we must add them to the regular expression.

@lmarcoruiz: 
This seems to be the solution for the ItemStructure and also a pattern we can reuse for the SysPeriod columns of all tables.
Can you look over the FolderAccess and FolderAccessTest if the adaptions to the usage of ItemStructure instead of Object and String is correct there?

### DIFF
--- a/jooq-pq/pom.xml
+++ b/jooq-pq/pom.xml
@@ -103,7 +103,7 @@
                                     <!-- Add a Java regular expression matching fully-qualified columns. Use the pipe to separate several expressions.
 
                                          If provided, both "includeExpressions" and "includeTypes" must match. -->
-                                    <expression>.*other_details.*</expression>
+                                    <expression>^(other_)?details$</expression>
                                     <!-- TODO tag name was rolled back to older version here, needs to be "includeExpressions" with newer jooq version again -->
 
                                     <!-- Add a Java regular expression matching data types to be forced to
@@ -121,6 +121,7 @@
                                     <types>.*</types>
                                     <!-- TODO tag name was rolled back to older version here, needs to be "includeTypes" with newer jooq version again -->
                                 </forcedType>
+
                             </forcedTypes>
                         </database>
                         <generate>
--- a/service/src/main/java/org/ehrbase/dao/access/jooq/FolderAccess.java
+++ b/service/src/main/java/org/ehrbase/dao/access/jooq/FolderAccess.java
@@ -137,7 +137,7 @@ public class FolderAccess extends DataAccess implements I_FolderAccess, Comparab
             }
 
             folderRecord.setSysPeriod(PGObjectParser.parseSysPeriod(folderRecord.getSysPeriod()));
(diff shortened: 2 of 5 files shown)


jooq-pq/pom.xml
```diff
@@ -103,7 +103,7 @@
                                     <!-- Add a Java regular expression matching fully-qualified columns. Use the pipe to separate several expressions.
 
                                          If provided, both "includeExpressions" and "includeTypes" must match. -->
-                                    <expression>.*other_details.*</expression>
+                                    <expression>^(other_)?details$</expression>
```
jakesmolka: Cool that you saw the problem with `audit_details` and avoided it :)

service/src/test/java/org/ehrbase/dao/access/jooq/FolderAccessTest.java
```diff
@@ -363,7 +363,22 @@ public void shouldUpdateExistingFolder() throws Exception {
         fa2.setFolderName("modifiedName");
         fa2.setFolderNArchetypeNodeId("modifiedArchetypeNodeId");
         fa2.setIsFolderActive(false);
-        fa2.setFolderDetails(DSL.field(DSL.val("{\"s\": \"modifiedValue\"}") + "::jsonb"));
+        ItemStructure is = new ItemStructure() {
```
jakesmolka: `ItemStructure` is an abstract class, right? From the openEHR RM perspective it is, see: https://specifications.openehr.org/releases/RM/latest/data_structures.html#_item_structure_class

So when you always assign a list style structure here, you could also directly create an non-abstract `ItemList` archie class instance. This would allow a leaner construction, too: https://github.com/openEHR/archie/blob/master/openehr-rm/src/main/java/com/nedap/archie/rm/datastructures/ItemList.java#L36

service/src/main/java/org/ehrbase/dao/access/jooq/FolderAccess.java
```diff
@@ -496,10 +496,10 @@ public static I_FolderAccess buildPlainFolderAccess(final  I_DomainAccess domain
         }
         //pstmt.setObject(11, jsonObject);
 
-        if (folder.getDetails() != null) {
+        /* if (folder.getDetails() != null) {
```
jakesmolka: Just taking a look at the snippet I'm not sure what happens here, but formally `details` in a Folder can be empty/null, see: https://specifications.openehr.org/releases/RM/latest/common.html#_folder_class (just as a note)

PR #51, opened 2019-11-24
#1 remove dependencies on child modules from parent pom

serefarikan:
can't see why parent pom lists child modules as dependencies.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -32,6 +32,7 @@
         <dependency>
             <groupId>org.ehrbase.openehr</groupId>
             <artifactId>serialisation</artifactId>
+            <version>${project.version}</version>
         </dependency>
         <dependency>
             <groupId>org.springframework.boot</groupId>
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -71,22 +71,27 @@
         <dependency>
             <groupId>org.ehrbase.openehr</groupId>
             <artifactId>service</artifactId>
+            <version>${project.version}</version>
         </dependency>
         <dependency>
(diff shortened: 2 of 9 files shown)


Geisterfalle:
Hi Seref,

I will ask @stefanspiska to take a look. I think there was a good reason, but I cannot remember. 

Cheers,

Birger

stefanspiska:
The parent pom in it self dos not include any dependencies it only sets via maven dependencies management the versions. Thus I would like to leave them there unless there is  good reason not to do it.

serefarikan:
@stefanspiska note that every dependency listed in the parent pom is listed again in the child poms, except without their versions. So in effect, children are only inheriting project version no (which is a property declared in the parent pom) and actually are overriding the dependency listed in the parent pom. 

The build works as it is, but every project that inherits the parent is also inheriting all the dependencies even though it is not needed. Inheriting the 3rd party deps in the parent pom is fine (as long as they're dependencies common to all child modules) but parent being dependant on its children would make me uncomfortable. 
I think you already have what you want in place with project version property inheritance, but it is up to your team of course.

stefanspiska:
No, that is not how that works 
see https://stackoverflow.com/questions/2619598/differences-between-dependencymanagement-and-dependencies-in-maven

elements in <dependencyManagement>  will not be automatically added to all sup-poms.
It only fixes the version wich will be included if you explicitly add the dependency to child pom.

serefarikan:
ah, yes, I forgot this difference. my point becomes subjective in this case and I don't want to insist. I'll close this issue.

Issue #61, opened 2019-12-02
Getting a version of a composition leaks db connection


grantforrester-nes:
After updating a composition, when getting a version of a composition, the database connection is not released.  Repeated use of this API causes db connection starvation.

Seen in v0.9.0.

**Steps To Reproduce**
1. POST a new composition to `ehrbase/rest/openehr/v1/ehr/{ehr_id}/composition`.
2. PUT an updated composition to `ehrbase/rest/openehr/v1/ehr/{ehr_id}/composition/{composition_id}`
2. Repeatedly GET the composition using the _versioned-composition-id_ `ehrbase/rest/openehr/v1/ehr/{ehr_id}/composition/{composition_id}::local.ethercis.com::1`

**Expected Result**
`200 OK`
`200 OK`
...
... _forever_

**Actual Result**
`200 OK`
...
... _8 more 200 OK responses then on 9th request_ ...
```
500 Internal Server Error
{
    "error": "Internal Server Error: org.jooq.exception.DataAccessException: Error getting connection from data source org.apache.commons.dbcp2.BasicDataSource@4833deb4",
    "status": "Internal Server Error"
}
```

grantforrester-nes:
I have attached PR #62 that fixes this issue.

Note this fix was also part of a PR that was pushed to Ethercis ([PR 22](https://github.com/ethercis/ehrservice/pull/22))

chevalleyc:
The PR changes are now integrated into the main codebase. Many thanks for the contribution!

grantforrester-nes:
Thanks 👍

PR #62, opened 2019-12-02
Get composition version connection leak fix

grantforrester-nes:
See #61


chevalleyc:
I have integrated the changes into https://github.com/ehrbase/ehrbase/tree/feature/62_get_composition_version_connection_leak_fix so this goes mainstream with CI checking.

PR #70, opened 2019-12-05
fix #68 by exporting cluster/name/value

### LINKED ISSUE #68: SQL syntax error when contributing (valid?) JSON
This may be related to #67 but I'm not 100% sure.

Currently I'm switching from XML to JSON and unfortunately ran into a few issues.
At first I thought that there might be an issue with my JSON document but it looks fine. 

So I did the following: Generated an XML composition and uploaded it to the Server. to get a valid JSON document I set the following headers:

- Prefer: return=representation
- Accept: application/json

Then I took the returned JSON document and tried to upload it. It returned the following error:
```
<?xml version='1.0' encoding='UTF-8'?>
<Map>
  <error>Internal Server Error: java.lang.IllegalArgumentException: Could not commit content:org.springframework.jdbc.BadSqlGrammarException: Access database using Jooq; bad SQL grammar [insert into "ehr"."containment" ("comp_id", "label", "path") values (?, 'openEHR_EHR_COMPOSITION_report_result_v1.openEHR_EHR_OBSERVATION_laboratory_test_result_v1.openEHR_EHR_CLUSTER_genetic_variant_v0]'::ltree, ?)]; nested exception is org.postgresql.util.PSQLException: ERROR: syntax error at position 129
  Position: 74</error>
  <status>Internal Server Error</status>
</Map>
```

I tried the same with another template and it worked fine, so it maybe has something to do with my template.

In the attachments:
- Template (*.opt and *.oet)
- XML Document I successfully uploaded
- JSON document that I tried to upload

[upload-json.txt](https://github.com/ehrbase/ehrbase/files/3917291/upload-json.txt)
[upload-xml.txt](https://github.com/ehrbase/ehrbase/files/3917292/upload-xml.txt)
[MTB-oet.txt](https://github.com/ehrbase/ehrbase/files/3917294/MTB-oet.txt)
[MTB-opt.txt](https://github.com/ehrbase/ehrbase/files/3917299/MTB-opt.txt)

nr23730:
This PR fixes the missing representation of name/value in clusters when using the raw format (XML and JSON).

### DIFF
--- a/serialisation/src/main/java/org/ehrbase/ehr/encode/wrappers/json/writer/translator_db2raw/ArrayClosure.java
+++ b/serialisation/src/main/java/org/ehrbase/ehr/encode/wrappers/json/writer/translator_db2raw/ArrayClosure.java
@@ -28,11 +28,13 @@ public class ArrayClosure {
     JsonWriter writer;
     String parentItemsArchetypeNodeId = null;
     String parentItemsType = null;
+    String parentItemsName = null;
 
-    public ArrayClosure(JsonWriter writer, String parentItemsArchetypeNodeId, String parentItemsType) {
+    public ArrayClosure(JsonWriter writer, String parentItemsArchetypeNodeId, String parentItemsType, String parentItemsName) {
         this.writer = writer;
         this.parentItemsArchetypeNodeId = parentItemsArchetypeNodeId;
         this.parentItemsType = parentItemsType;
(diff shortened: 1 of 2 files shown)


chevalleyc:
These changes have been integrated into https://github.com/ehrbase/ehrbase/pull/82 (for use with the CI framework). Many thanks for the contribution!

PR #71, opened 2019-12-06
Feature/46 validation and cached openehr terminology

chevalleyc:
This PR is documented in https://wiki.vitagroup.ag/display/ETHERCIS/Validation+Using+a+Terminology?src=contextnavpagetreemode

PR #79, opened 2019-12-12
Feature/46 terminology validation

chevalleyc:
After quite a lot of pushes (my Git changelist seems to have a lot of problem...) and robot test suite fixes, this PR replaces https://github.com/ehrbase/ehrbase/pull/71


api/pom.xml
```diff
@@ -61,6 +61,11 @@
             <groupId>commons-lang</groupId>
             <artifactId>commons-lang</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.ehrbase.openehr</groupId>
+            <artifactId>terminology</artifactId>
+            <version>${project.version}</version>
```
stefanspiska: The current pattern is to add the version in the dependence management of the main.pom such that in the module.poms not version is needed.

service/src/test/java/org/ehrbase/service/ValidationServiceTest.java
```diff
@@ -39,21 +42,26 @@
     public CacheRule cacheRule = new CacheRule();
 
     @Test
-    public void testCheckComposition() throws Exception {
+    public void testCheckComposition(){
+        try {
+            KnowledgeCacheService knowledgeCacheService = KnowledgeCacheServiceTest.buildKnowledgeCache(testFolder, cacheRule);
+            knowledgeCacheService.addOperationalTemplate(IOUtils.toByteArray(Files.newInputStream(Paths.get("./src/test/resources/knowledge/RIPPLE-ConformanceTest.opt"))));
```
stefanspiska: Please Use OperationalTemplateTestData.RIPPLE_CONFORMANCE_TEST.getStream()

terminology/pom.xml
```diff
@@ -0,0 +1,32 @@
...
+        <version>0.9.0</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>terminology</artifactId>
+    <dependencies>
+        <dependency>
+            <groupId>com.github.openEHR.archie</groupId>
+            <artifactId>openehr-rm</artifactId>
+            <version>9d65960</version>
```
stefanspiska: You should'n need the version since it is defined in the main.pom

terminology/pom.xml
```diff
@@ -0,0 +1,32 @@
...
+        </dependency>
+        <!-- Test-->
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
```
stefanspiska: please move this above the  comment

validation/pom.xml
```diff
@@ -55,5 +55,20 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.ehrbase.openehr</groupId>
+            <artifactId>serialisation</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.ehrbase.openehr</groupId>
+            <artifactId>terminology</artifactId>
+            <version>${project.version}</version>
```
stefanspiska: Version should imo be set in the main.pom. see above

validation/src/main/java/org/ehrbase/validation/terminology/I_ItemStructureVisitor.java
```diff
@@ -0,0 +1,31 @@
...
+import com.nedap.archie.rm.composition.Composition;
+import com.nedap.archie.rm.datastructures.ItemStructure;
+
+public interface I_ItemStructureVisitor {
+
+    void validate(Composition composition) throws Throwable;
+
+    void validate(ItemStructure itemStructure) throws Throwable;
+
+    void validate(Locatable locatable) throws Throwable;
```
stefanspiska: Please fix what exactly can be thrown.

validation/src/main/java/org/ehrbase/validation/terminology/ItemStructureVisitor.java
```diff
@@ -0,0 +1,424 @@
...
+public class ItemStructureVisitor implements I_ItemStructureVisitor {
+
+    protected static Logger log = LoggerFactory.getLogger(ItemStructureVisitor.class);
+    private int elementOccurrences = 0; //for statistics and testing
+    private ItemValidator itemValidator = new ItemValidator();
+    private LocalizedTerminologies localizedTerminologies;
+    private AttributeCodesetMapping codesetMapping;
+    private String itemStructureLanguage = "en"; //if a composition, the language can be found in the structure
+
+    public ItemStructureVisitor(LocalizedTerminologies localizedTerminologies) throws NoSuchMethodException, IllegalAccessException, ClassNotFoundException {
```
stefanspiska: I would be better to catch this Exceptions where there are occurring and wrap them in a Validation Exception. 
I also think this wrapper can be a RuntimeException.

validation/src/main/java/org/ehrbase/validation/terminology/ItemValidator.java
```diff
@@ -0,0 +1,119 @@
...
+    private Map<String, ValidationHandler> validationRegistryList;
+
+    private static final String VALIDATOR_PACKAGE = "org.ehrbase.validation.terminology.validator";
+
+    public ItemValidator() {
+        validationRegistryList = new HashMap<>();
+   }
+
+   public ItemValidator add(Class rmObjectClazz) throws NoSuchMethodException, IllegalAccessException, ClassNotFoundException {
+       Class validationHandlerClass = Class.forName(VALIDATOR_PACKAGE+"."+rmObjectClazz.getSimpleName());
```
stefanspiska: Suggestion:
Instead of using names to map  RMclasses to it corresponding TerminologyCheck it would be better if each TerminologyCheck  has a getRMClass which returns the RMclass it is for and use that as matching.

validation/src/main/java/org/ehrbase/validation/terminology/ItemStructureVisitor.java
chevalleyc: I had opened a ticket for this as I believe this is a wider change https://github.com/ehrbase/project_management/issues/84

validation/src/main/java/org/ehrbase/validation/terminology/I_ItemStructureVisitor.java
chevalleyc: Fixing this issue is also related to ticket https://github.com/ehrbase/project_management/issues/84. This is a wider issue than this PR.

api/pom.xml
chevalleyc: fixed with the new commit

service/src/test/java/org/ehrbase/service/ValidationServiceTest.java
chevalleyc: fixed with the new commit

terminology/pom.xml
chevalleyc: fixed with the new commit

terminology/pom.xml
chevalleyc: fixed with the new commit

validation/pom.xml
chevalleyc: fixed with the new commit

validation/src/main/java/org/ehrbase/validation/terminology/I_ItemStructureVisitor.java
chevalleyc: to add to this, the latest commit does fix some of the above. However, a specific ticket/task should address the validation error (validation exception) globally.

validation/src/main/java/org/ehrbase/validation/terminology/ItemValidator.java
chevalleyc: fixed with the new commit

pom.xml
```diff
@@ -636,7 +646,7 @@
       <properties>
         <gpg.keyname>FAB0BE712374BFC3D26DA5B43A7041C227272DD8</gpg.keyname>
         <gpg.executable>gpg2</gpg.executable>
-        <gpg.passphrase>${env.GPG_PASSPHRASE}</gpg.passphrase>
+        <!--<gpg.passphrase>${env.GPG_PASSPHRASE}</gpg.passphrase>-->
```
stefanspiska: I think this needs to not be commented out

chevalleyc: Oh I see.. this causes an error in my environment... any link to a documentation on this?

chevalleyc: Oh well. did uncommented this. Feels like black magic somehow as this definitely shouldn't result in an error in any environment unless proper guidelines are provided explaining the Why's and How's.

PR #86, opened 2019-12-17
added validation of other_context

chevalleyc:
This add the validation for composition/context/other_context


chevalleyc:
This is to fix https://github.com/ehrbase/project_management/issues/87

PR #89, opened 2019-12-19
ehrbase/project_management#98 added ACTION examples and updated AQL test suite data

ppazos:
@testautomation this includes the action compositions and the queries were updated too.


testautomation:
:heavy_check_mark:

PR #91, opened 2019-12-19
19 integration test improvements

testautomation:
This PR mainly consolidates all the different start ... request session keywords from several resource files into one keyword prepare new request session in one resource file --> _resources/keywords/generic_keywords.robot and updates all test cases and keywords that were impacted by this change.


testautomation:
> just wondering if we need to force the accept header in all requests

@ppazos  It would be easier not to force it always but it was the result of different behaviour between EHRbase and EHRSCAPE - some request did not worked correctly when accept or content were not specified.

PR #96, opened 2020-01-08
ehrbase/project_management#105 added missing order

ppazos:
@testautomation please check this solution for ehrbase/project_management#105


testautomation:
✔️

Issue #67, opened 2019-12-03
Inconsistent casting between XML and JSON


nr23730:
I just discovered some inconsistencies between XML and JSON. To me it looks like the conversion (XML -> JSON) already already removes invalid items so that the JSON parsing completes successfully.

When I submit the following item in a contribution (I know that `1,2,3,4` is not a valid magnitude!)
```
<items archetype_node_id="at0067" xsi:type="ELEMENT">
	<name>
		<value>Strand bias ratio</value>
	</name>
	<value xsi:type="DV_QUANTITY">
		<magnitude>1,2,3,4</magnitude>
		<units>1</units>
	</value>
</items>
```
it resulsts in the following response (Prefer: return=representation, Accept: application/json):
```
{
	"_type": "ELEMENT",
	"name": {
		"_type": "DV_TEXT",
		"value": "Strand bias ratio"
	},
	"value": {
		"_type": "DV_QUANTITY",
		"units": "1"
	},
	"archetype_node_id": "at0067"
}
```

So as a result the invalid magnitude will be removed.

But when I try to upload an equivalent representation in JSON:
```
{
	"archetype_node_id": "at0067",
	"_type": "ELEMENT",
	"name": {
		"_type": "DV_TEXT",
		"value": "Strand bias ratio"
	},
	"value": {
		"_type": "DV_QUANTITY",
		"magnitude": "1,2,3,4",
		"units": "1"
	}
}
```
I get the following error message (Error 500):
```
Cannot deserialize value of type `java.lang.Double` from String "1,2,3,4": not a valid Double value
```

I'm not sure which behaviour is intented but I think that XML and JSON should both either remove such invalid items or throw the same error message.

ppazos:
@nr23730 there is no 1 to 1 correspondence between the formats. I guess the differences in the validation are because the validators are different, one is against the XML schema and the other against the JSON schema, which are different and might validate different things in a different way. Also I think the JSON schema still needs some fixes.

Besides that, the API should not return 500 error in any case, not sure if that was your case, but also not sure if you were using the API or the internal code.

Geisterfalle:
@ppazos The validation for the constraints should not be different between the formats as  there is only one validator being executed on the RM instances and not on the serialization format.

ppazos:
@Geisterfalle thought the validation was using archie and the schemas. Also, from the issue description, it is not clear what is the context of execution to get the error. Do we have any unit test checking this?

stefanspiska:
[...]

PR #99, opened 2020-01-09
check (some) mandatory attribute value and fix incorrect exception re…

chevalleyc:


chevalleyc:
This is to fix CR https://github.com/ehrbase/ehrbase/issues/67

PR #100, opened 2020-01-10
Feature/76 delete directory rest

axelsiebert:
Added the versionUid helper class for future validation on versionUid strings but it is not used yet.

Maybe have a look on the .gitignore for VisualStudio Code since this vonfig would allow the launch.json file to be committed. Maybe it is not intended.

Known Issue:

- There is an issue with the version number detection which always returns 1.

### DIFF
--- a/service/src/main/java/org/ehrbase/service/FolderServiceImp.java
+++ b/service/src/main/java/org/ehrbase/service/FolderServiceImp.java
@@ -24,6 +24,7 @@ import org.ehrbase.api.definitions.ServerConfig;
 import org.ehrbase.api.definitions.StructuredString;
 import org.ehrbase.api.definitions.StructuredStringFormat;
 import org.ehrbase.api.dto.FolderDto;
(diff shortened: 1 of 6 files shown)


service/src/main/java/org/ehrbase/service/FolderServiceImp.java
```diff
@@ -175,8 +176,13 @@ public UUID create(UUID ehrId, Folder content) {
...
     public LocalDateTime delete(UUID folderId) {
 
-        // TODO implement logic
-        return LocalDateTime.now();
+        I_FolderAccess folderAccess = I_FolderAccess.retrieveInstanceForExistingFolder(getDataAccess(), folderId);
+
+        if (folderAccess.delete() > 0){
+            return LocalDateTime.now();
+        } else {
+            throw new InternalServerException("Folder " + folderId + " could not be deleted");
```
lmarcoruiz: should we specify the reason causing the lack of deletion??

PR #101, opened 2020-01-13
json blocks as strings, correct archetype node and names id in contents

chevalleyc:
This fixes the issue in https://github.com/ehrbase/project_management/issues/85 as well as a fix for correctly passing archetype node ids in content children. Solve warnings in this code. Added @SuppressWarnings("unchecked") for casting on LinkedTreeMap with generics since this cannot be solved at compile time.


chevalleyc:
created a new PR with the changes https://github.com/ehrbase/ehrbase/pull/115

PR #102, opened 2020-01-14
added resolution of ehr/system_id in AQL

chevalleyc:
Fix issue #110


chevalleyc:
Modified to return the right system_id/value (was settings...)

PR #113, opened 2020-01-24
use openehr instead of openEhr as terminology_id

nr23730:
According to https://specifications.openehr.org/releases/TERM/latest/SupportTerminology.html the correct term for the support terminology is `openehr` and not `openEhr`.

Contributions were only accepted with the `openehr` term while the output always produced `openEhr`.


ppazos:
@testautomation the JSON example fixed in this PR doesn't seem to be valid at all, but having an old format, I don't think it is used by any test, can you confirm? It's:

minimal_observation.composition.participations.extdatetimes.ehrscape.json

PR #118, opened 2020-01-28
ehrbase/project_management#51

jakesmolka:
Fixes POST composition response when template from payload does not
exist.

@testautomation  can you check if the modification of the robot file is correct?
https://github.com/ehrbase/ehrbase/compare/feature/51_bug_posting_compo_wo_template?expand=1#diff-4a3e54850144b4ac6e32e5901aa8e944L117

fixes ehrbase/project_management#51

### DIFF
--- a/api/src/main/java/org/ehrbase/api/exception/UnprocessableEntityException.java
+++ b/api/src/main/java/org/ehrbase/api/exception/UnprocessableEntityException.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright (c) 2019 Vitasystems GmbH and Jake Smolka (Hannover Medical School).
+ *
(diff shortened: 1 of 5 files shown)


testautomation:
> can you check if the modification of the robot file is correct?

looks good 👍 


We have another test that is tagged with the same issue ID ehrbase/project_management#51:
CONTRIBUTION.C.1 COMMIT CONTRIBUTION.d) With multiple version composistions

@jakesmolka  Can you please check if the test should also check for status code 422 instead of 400?

jakesmolka:
@testautomation This test you mentioned should check for status code 400. So it is correct!? 
And so there's no further connection to this PR, right? I would merge then.

PR #120, opened 2020-01-29
Feature/82 versioned object poc

jakesmolka:
Original task was to research `VERSIONED_OBJECT` capabilities with our code base. This PR went from research to actual implementation of four REST endpoints, from access layer to presentation layer.

Endpoints: see the four `VERSIONED_EHR_STATUS` ones here https://specifications.openehr.org/releases/ITS-REST/latest/ehr.html#ehr_status-versioned_ehr_status

`OriginalVersion*`, `RevisionHistory*` and `VersionedObjectResponseData` to map the service output to the specified format.

Controller for the endpoints: `OpenehrVersionedEhrStatusController`

`EhrServiceImp` got new methods to generate new necessary data and invoke access layer methods.
- Specifically important is generation of the `OriginalVersion<>` wrapper, `VersionedEhrStatus` and `RevisionHistory` to comply with the endpoints (especially the Attestations included in RevisionHistoryItem needed many new changes).

Access layer:

- `EhrAccess`
    - Has new associated `StatusAcess`, instead of embedded `StatusRecord` so all status methods can be maintained in one place (still needs refactoring).
    - Some new methods, like one to determine the time of the oldest Status version
- `StatusAccess`
    - Basically whole implementation of CRUD like methods.
    - Includes internal handling of associated audit and contribution if appropriate.
- `AttestationAccess`
    - Has associated `AuditDetailsAccess` to model Attestation being a subclass of AuditDetails
    - Only reading/retrieval functionality for now
    - Retrieval of all attestations related to the given `attestation_ref` (see below)
- And some small changes to other Access classes

DB schema changes:
- "obfuscation" table `attestation_ref` to allow pointing from version objects of any kind to attestations without breaking the internal postgres type checking
- new `attestation_ref` field for all objects that can be/have `ORIGINAL_VERSION` and there can have associated attestations.
- Removal of versioning (via temporal tables) of `audit_details` to improve maintainability

closes ehrbase/project_management#82


chevalleyc:
Unless I miss something, did you implement any unit tests to illustrate the behaviour? I couldn't find any. Thanks!

chevalleyc:
I have a some questions/comments:

1. What is the relationship between `version_uid` f.e. `preceding_version_uid` and history table? In other words, whenever a composition is amended, how does this materialize at DB level?
2. I've noticed the state of the contribution is not saved, does this mean that whenever we want to query the history of changes for a composition (e.g. creation, amend, etc.) we are getting a list of items without state?
3. I think some UML (Use Case and Sequence Diagram) could help a lot to understand this implementation.

NB. My concern is to understand the impact of the above to implement AQL querying on RM object history (f.e. TIME WINDOW)

jakesmolka:
@chevalleyc 
1. I haven't introduced those columns nor worked on composition in this PR. 
So in a general scope: Amendment is handled about the same as modifications for now. A new version will be created, i.e. the current entry from, for instance, the `composition` table will be moved to the `composition_history` table, while the new one will be written into `composition`, with the same ID. But since you know that, I'm not sure what your question is.
The `preceding_version_uid` depends on the use case. In a REST request with the `If-Match` header it semantically equals the current version. While in other cases it means "second latest", which can be queried by looking into the `*_history` table.
2. Where? That wouldn't be intentional. In my tests there were some cases where the state wasn't saved (older tech dept), but in general it was.
3. Great idea! Do you think about this PR's scope or the whole access layer?

chevalleyc:
Many thanks for the explanation Jakes!

PR #121, opened 2020-01-30
fix ehr.js_dv_date_time

chevalleyc:
Noticed that plpgsql function ehr.ehr.js_dv_date_time required to specify argument as `TIMESTAMP WITHOUT TIME ZONE` to be properly interpreted


chevalleyc:
I've changed this function following our discussion last Friday, this one returns:
```
2019-01-15 01:36:49.294+07
```
for this arguments:
```
select timezone('Asia/Bangkok', '2019-01-15 01:36:49.294'::timestamp)
```

PR #124, opened 2020-02-04
Feture/some smal fixes

stefanspiska:


service/src/main/java/org/ehrbase/dao/access/jooq/FolderAccess.java
```diff
@@ -142,9 +142,8 @@ private Boolean update(final Timestamp transactionTime,
         // Copy into new instance and attach to DB context.
         // The new instance is required to store the new record with a new ID
         FolderRecord updatedFolderRecord = new FolderRecord();
-        if (rootFolder) {
-            updatedFolderRecord.setId(this.getFolderId());
-        }
+        updatedFolderRecord.setId(this.getFolderId());
```
lmarcoruiz: why do you want to maintain the id of old subfolders that are actually complete  new folders that override the old ones?

service/src/main/java/org/ehrbase/dao/access/util/FolderUtils.java
```diff
@@ -32,6 +34,7 @@ public static void updateFolder(Folder update, I_FolderAccess target) {
             target.getItems().clear();
             target.getItems().addAll(update.getItems());
         }
+        target.setFolderId(UUID.fromString(update.getUid().getValue()));
```
axelsiebert: At this place the update Folder object does not contain an id. The id will be given by the database during commit().

Thus I get a NullPointerException here when trying to update a folder with a basic set.

Data for create via POST:
```json
{
    "_type": "FOLDER",
    "name": {
        "_type": "DV_TEXT",
        "value": "Root folder"
    },
    "archetype_node_id": "at[0001]",
    "folders": [
    	{
    		"_type": "FOLDER",
    		"name": {
    			"_type": "DV_TEXT",
    			"value": "Sub folder"
    		},
    		"archetype_node_id": "at[0001]"
    	}
    ]
}
```

And data for update via PUT:
```json
{
    "_type": "FOLDER",
    "name": {
        "_type": "DV_TEXT",
        "value": "Updated Root name"
    },
    "archetype_node_id": "at[0001]",
    "folders": [
        {
            "_type": "FOLDER",
            "name": {
                "_type": "DV_TEXT",
                "value": "Updated Sub folder"
            },
            "archetype_node_id": "at[0001]"
        }
    ]
}
```

PR #125, opened 2020-02-05
Feature/84 clean validation exceptions

chevalleyc:


axelsiebert:
Please ahve a look since it is blocking issue https://github.com/ehrbase/project_management/issues/61

serialisation/src/main/java/org/ehrbase/serialisation/CanonicalJson.java
```diff
@@ -90,7 +90,7 @@ public String marshal(RMObject rmObject) {
         try {
             return JacksonUtil.getObjectMapper().readValue(value, clazz);
         } catch (IOException e) {
-            throw new UnmarshalException(e.getMessage(), e);
```
stefanspiska: Can you life that as UnmarshalException since that is the internal Exception of the serialisation  module

rest-openehr/src/main/java/org/ehrbase/rest/openehr/controller/BaseController.java
```diff
@@ -87,8 +87,8 @@
         return metaMap;
     }
 
-    protected String getBaseEnvLinkURL() {
```
stefanspiska: I would like to leave them as protected if someday we move the split the controller in deferent packages

stefanspiska:
replaced by https://github.com/ehrbase/ehrbase/pull/146

PR #140, opened 2020-02-17
Provider name error

robert-hunter-nes:
I've added a PR to fix an issue I have seen when saving a composition that contains a provider

* Updating CompositionSerializer to handle the PartyIdentified name field as a string rather than DvText
* Updating diadem.xml to include provider on an evaluation so the RawJsonTest unmarshall test confirms there is no errors deserialising the provider details

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,12 @@ cd tests
  2. Browse to Swagger UI --> http://localhost:8080/ehrbase/swagger-ui.html
 
 
+## Docker
+
+To create a Docker image
+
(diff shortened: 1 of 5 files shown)


robert-hunter-nes:
Please don't review this yet, I need to sort out the file cahnges as I only made 2 but I have extra changes in this PR. I'll update once I have resolved it.

robert-hunter-nes:
Closing this PR as there are unrequired changes in it, planning to open another one

PR #150, opened 2020-02-25
Fix test for Duration and OffsetDateTime

chevalleyc:
~This fixes this specific issue~

fixes https://github.com/ehrbase/project_management/issues/61

### DIFF
--- a/validation/src/main/java/org/ehrbase/validation/constraints/wrappers/CDuration.java
+++ b/validation/src/main/java/org/ehrbase/validation/constraints/wrappers/CDuration.java
@@ -37,15 +37,12 @@ import java.util.Map;
  */
 public class CDuration extends CConstraint implements I_CTypeValidate {
 
-    CDuration(Map<String, Map<String, String>> localTerminologyLookup) {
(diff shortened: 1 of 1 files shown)


validation/src/main/java/org/ehrbase/validation/constraints/wrappers/CDuration.java
```diff
@@ -37,15 +37,12 @@
...
 public class CDuration extends CConstraint implements I_CTypeValidate {
 
-    CDuration(Map<String, Map<String, String>> localTerminologyLookup) {
+    protected CDuration(Map<String, Map<String, String>> localTerminologyLookup) {
         super(localTerminologyLookup);
     }
 
     @Override
-    public void validate(String path, Object aValue, CPRIMITIVE cprimitive) throws IllegalArgumentException {
-        if (!(aValue instanceof OffsetDateTime))
```
ppazos: @chevalleyc isn't the type check needed here?

chevalleyc: This is now done by Archie (if the string cannot be parsed, it is null and rejected by the validator)

PR #162, opened 2020-02-28
Feature/169 implement fhir ts client

lmarcoruiz:
Develops first implementation clossing the loop with a terminology service (ontoserver FHIR)n and generating the SQL code for the MATCHES section with its response.


service/src/test/java/org/ehrbase/aql/compiler/InvokeVisitorTest.java
```diff
@@ -30,121 +30,22 @@
 
 public class InvokeVisitorTest {
```
chevalleyc: I would suggest to implement a standalone unit test that doesn't require an external service. F.e. using MockServer:

- https://www.mock-server.com/
- https://www.baeldung.com/mockserver

This would allow to enrich the test suite as required (authentication, other query types or formats etc.)

service/src/main/java/org/ehrbase/aql/compiler/tsclient/fhir/FhirTerminologyServerImpl.java
```diff
@@ -0,0 +1,102 @@
...
+		RestTemplate rest = new RestTemplate();
+		HttpHeaders headers = new HttpHeaders();
+		headers.set("accept","application/fhir+json");
+		HttpEntity<String> entity =  new HttpEntity<String>(headers);
+		ResponseEntity<String> responseEntity = rest.exchange(valueSetId,
+				HttpMethod.GET,
+				entity,
+				String.class);
+		String response = responseEntity.getBody();
+		String jsonCodePath = "$[\"expansion\"][\"contains\"][*][\"code\"]";
```
chevalleyc: is jsonCodePath a "universal" pattern? If not (e.g. depends on service), I would suggest to externalize it allowing to be maintained without requiring further coding.

lmarcoruiz: yes. since ValueSets are standardized the json path is stable across ValueSets.

service/src/test/java/org/ehrbase/aql/compiler/InvokeVisitorTest.java
lmarcoruiz: yes. we have this in netx sprint task as part of the exception management. I will develop the server mockup and mock exceptions as well.

PR #173, opened 2020-03-05
added explicit UTF-8 encoding when reading to byte

Geisterfalle:
## Changes

Added explicit conversion to UTF-8.

- First change

## Related issue

Fixes https://github.com/ehrbase/project_management/issues/112

### DIFF
--- a/service/src/main/java/org/ehrbase/service/TemplateFileStorageService.java
+++ b/service/src/main/java/org/ehrbase/service/TemplateFileStorageService.java
@@ -26,10 +26,8 @@ import org.springframework.beans.factory.annotation.Autowired;
 
 import javax.annotation.PostConstruct;
 import javax.xml.namespace.QName;
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.IOException;
-import java.io.InputStream;
+import java.io.*;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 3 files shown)


service/src/main/java/org/ehrbase/service/TemplateFileStorageService.java
```diff
@@ -123,7 +121,7 @@ void setOptPath(String optPath) {
     public void storeTemplate(OPERATIONALTEMPLATE template) {
         XmlOptions opts = new XmlOptions();
         opts.setSaveSyntheticDocumentElement(new QName("http://schemas.openehr.org/v1", "template"));
-        saveTemplateFile(template.getTemplateId().getValue(), template.xmlText(opts).getBytes());
+            saveTemplateFile(template.getTemplateId().getValue(), template.xmlText(opts).getBytes(StandardCharsets.UTF_8));
```
ppazos: @Geisterfalle please fix indentation

Geisterfalle:
Thanks Pablo. You can use any template and when there are no JVM parameters or environment variables provided, the upload fails on system were UTF-8 is not standard encoding (windows, some Linux distributions).

Geisterfalle:
@ppazos should be good now, please have a look

ppazos:
@Geisterfalle I'm not running the system with an encoding parameter, in fact I never did, and didn't got errors when uploading templates on my environment, so I can't reproduce since I don't have the same env.

PR #180, opened 2020-03-10
Feature/147 ehr status version uid

jakesmolka:
## Changes

Fixes STATUS uid. Now correctly is a **version_object_uid**, instead of *versioned-...*.

## Related issue

closes https://github.com/ehrbase/project_management/issues/147


jakesmolka:
Blocked by @axelsiebert's issue https://github.com/ehrbase/project_management/issues/148 because of overlap. Wait until it is merged and see which changes of this PR needs to be reverted.

rest-openehr/src/main/java/org/ehrbase/rest/openehr/controller/OpenehrEhrStatusController.java
```diff
@@ -182,10 +182,11 @@ public OpenehrEhrStatusController(EhrService ehrService) {
 
         List<String> headerList = Arrays.asList(CONTENT_TYPE, LOCATION, ETAG, LAST_MODIFIED);   // whatever is required by REST spec
         Optional<InternalResponse<EhrStatusResponseData>> respData;   // variable to overload with more specific object if requested
+        UUID statusUid = UUID.fromString(status.getUid().getValue().split("::")[0]);
```
ppazos: @jakesmolka note in the spec, the OBJECT_VERSION_ID is

uid: UID
system_id: UID
version_tree_id: VERSION_TREE_ID

UID is abstract, could be UUID, OID or INTERNET_ID

Here status.getUid() should be an OBJECT_VERSION_ID, so the first part is a UID, not a UUID. For instance the uid could be "2.16.234.234.4353.364" (an OID).

If we are not going to support OIDs I guess we need to document it.

REF: https://specifications.openehr.org/releases/BASE/latest/base_types.html#_overview_4

jakesmolka:
Going to merge this as the linked integration test now is passed successfully. But be aware that @ppazos note this needs to be processed and worked on, see: https://github.com/ehrbase/project_management/issues/203

PR #183, opened 2020-03-10
Feature/full folder uid handling

axelsiebert:
## Changes

- Add handling for version_uid from request path
- Add handling for version_uid from If-Match header
- Provide root folder uids as ObjectVersionId with version number
- Provide sub folder uids as HierObjectId without version number
- Integrate VersionUidHelper for using version_uid over all instances of request handling
- Extent PreconditionFailedException for adding ETag and Location headers to error response

## Related issue

Fixes https://github.com/ehrbase/project_management/issues/148


api/src/main/java/org/ehrbase/api/service/FolderService.java
```diff
@@ -79,9 +79,8 @@
      * Updates a target folder entry identified by the given folderId with new
      * content. The content string will be serialized from the given source
      * format.
-     * TODO: Copy from CompositionService. Must be designed for folder
      *
-     * @param folderId - Id of the target folder
+     * @param folderId - Full version_uid for folder including system id and version
```
ppazos: @axelsiebert the correct type would be OBJECT_VERSION_ID https://specifications.openehr.org/releases/BASE/latest/base_types.html#_object_version_id_class

api/src/main/java/org/ehrbase/api/util/VersionUidHelper.java
```diff
@@ -35,6 +35,19 @@
                             VERSION_PATTERN.toString()
             );
 
+    private UUID uuid;
```
ppazos: @axelsiebert note in the openEHR model the OBJECT_VERSION_ID fields have different types:

uid: UID   // attribute is "uid" not "uuid"
system_id: UID
version_tree_id: VERSION_TREE_ID

UID is abstract, could be UUID, OID or INTERNET_ID

By assigning a UUID/String to the UID fields, and int to the VERSION_TREE_ID is constraining what is possible in the spec. For instance the uid could be "2.16.234.234.4353.364" (an OID), the system_id "org.x.y" (an INTERNET_ID), and the version_tree_id "2.1.3" (a VERSION_TREE_ID).

https://specifications.openehr.org/releases/BASE/latest/base_types.html#_overview_4

Note this class might already be implemented in the Java Ref Impl class and maybe in archie.

testautomation:
@axelsiebert  should we also reference https://github.com/ehrbase/project_management/issues/48 in this PR? I think it should also be closed by this PR.

CHANGELOG.md
```diff
@@ -11,10 +11,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - Update of AQL-Query test suite (see: https://github.com/ehrbase/ehrbase/pull/179)
 
 ### Fixed
 
+#### DIRECTORY
+
+- Directory IDs from input path or If-Match header must now be in version_uid format (see https://github.com/ehrbase/ehrbase/pull/183)
+- Folder IDs inside body are now parsed correctly (see: https://github.com/ehrbase/ehrbase/pull/183)
+
+#### Robot Tests
```
jakesmolka, line 30: Quick comment: I'm not a fan of those sub-headers. Would you argue we benefit from them? I think this is just one layer more of where stuff will be added wrong into the changelog ;) For instance, the linked #207 below isn't just a robot test update, instead it introduced the validation of EHR_STATUS object and fixed other bugs.

service/src/main/java/org/ehrbase/dao/access/interfaces/I_FolderAccess.java
```diff
@@ -94,6 +95,14 @@ static I_FolderAccess retrieveInstanceForExistingFolder(I_DomainAccess domainAcc
         return FolderHistoryAccess.retrieveInstanceForExistingFolder(domainAccess, folderId, timestamp);
     }
 
+    /**
+     * Creates a new directory object with a given structure and returns a valid Object_Version_Id containing the given
+     * system identifier and version part.
+     *
+     * @return Object_Version_Id for new root directory folder
+     */
+    ObjectVersionId create();
```
jakesmolka, line 104: Did you define this to circumvent the fix `UUID` return type for `commit(..)` set by the `I_SimpleCRUD` interface, right? Clever :) I was just trying to use `ObjectVersionId` in the EHR access layer and had the same problem.
In the end I'm wondering if it would make more sense to just add a general set of CRUD methods which are returning `ObjectVersionId` to the `I_SimpleCRUD` interface (or to add a separate interface class like `I_VersionedCRUD`). So it can be used consistently between all access classes (of versioned objects). But I think if it works, it is fine and okay for this case here.

PR #184, opened 2020-03-11
Feature/182 composition composer

chevalleyc:
## Changes

Added support of AQL query in the form of c/composer

## Related issue

closes: https://github.com/ehrbase/project_management/issues/182


Geisterfalle:
One more small thing:

When using the path inside SELECT:

` a/composer/external_ref/id/value`

I get the following error message:

```
{
  "error": "Unresolved party_identified external_ref attribute path:external_ref/id/value",
  "status": "Bad Request"
}
```

If I am not mistaken, this should be a valid path

chevalleyc:
I suggest to merge this anyway and to open a CR for the bugs you reported. This is necessary as the changes for CR #215 also require a new flyway migration which will collide with the one pushed with this CR. Thanks!

chevalleyc:
I should have added that the above issues will be resolved with CR #215 since the Composer resolver has to be revisited :-)

PR #198, opened 2020-03-23
181 further work fhir ts spring config

lmarcoruiz:
## Changes

*Makes properties Spring configurable for reading results from the external terminology server.

*Refactors to allow using Spring properties read from the application module.

## Related issue

Closes https://github.com/ehrbase/project_management/issues/209

Includes the implementation to read the response from the FHIR TS into DvCodedText


lmarcoruiz:
Hi @stefanspiska , thanks for the comments:-)
1-yes autowired was not in use there. It is deleted now.
2-I do not think that is a problem because lexical, syntactical and semantic analysis occur at different stages. So a client with AQL autocorrect can perform "online" lexical and  syntactic analysis (at client side), while the semantic analysis is later performed on the server side resolving the terminology invocation into the list of codes for the SQL statement. I think that will be more efficient also if we have several MATCHES clauses.
3-Agree, just waiting for the openEHR SEC consensus on this.

PR #199, opened 2020-03-23
Fixes POST without accept for 3 endpoints

jakesmolka:
Fixes default values and adds test for POST ehr, composition and
contribution.

## Changes

- Fixes POST without accept for 3 endpoints
  - Fixes default values and adds test for POST ehr, composition and
contribution.

More details on test changes:

POST ehr:
- existing `MF-001 - Create new EHR (w/o Prefer header)` already does check it. Renamed to `MF-001 - Create new EHR (w/o Prefer and Accept header)` to prevent accently refacotring away from that feature.

POST composition:
- created new `commit composition without accept header` keyword which basically is a copy of the normal one, but without Accept header.
- used in new `Main flow has existing COMPOSITION and works without accept header` test case

POST contribution:
- created new `POST /ehr/ehr_id/contribution without accept header` (... like above)
- used by newly created `commit CONTRIBUTION without accept header` (... like above)
- used in new `Main flow: successfully commit CONTRIBUTION with single valid VERSION<COMPOSITION> withtout accept header` test case

KW **prepare new request session** refactored:
- Requests can now be triggered w/ or w/o Accept/Content-Type headers
- JSON (default) - both (accept and content-type) set to application/json
- XML (both set to application/xml)
- no accept header (content-type=application/json)
- no content header
- no accept header xml (content-type=application/xml)
- no accept/content-type headers (other 'extra_headers' like IF-Match still possible)
- no headers

## Related issue

fixes https://github.com/ehrbase/project_management/issues/175


testautomation:
@jakesmolka  

1. I found an issue which requires me to refactor an existing keyword or add and additional keyword. FYI: `prepare request session` KW sets Content and Accept headers to application/json even if called w/o format argument (JSON).

image

PR #204, opened 2020-03-25
Feature/fix aql result

stefanspiska:
## Changes

some minor fixes to the aql result

- corrected _type value for context
- time format when directly selecting a field


stefanspiska:
added fix return defining_code as json

PR #207, opened 2020-03-31
Feature/193 other details validation ehrstatus attr

chevalleyc:
## Changes

This fixes multiple issues related to CR #193

Most important change is the new DB layout for table STATUS: added archetype_node_id and name column.

## Related issue

closes: https://github.com/ehrbase/project_management/issues/193

Edit by @jakesmolka:
fixes https://github.com/ehrbase/project_management/issues/157
fixes https://github.com/ehrbase/project_management/issues/158
fixes https://github.com/ehrbase/project_management/issues/159

### DIFF
--- a/api/src/main/java/org/ehrbase/api/service/ValidationService.java
+++ b/api/src/main/java/org/ehrbase/api/service/ValidationService.java
@@ -62,7 +62,7 @@ public interface ValidationService {
      * @param ehrStatus
      * @throws IllegalArgumentException
      */
-    void check(EhrStatus ehrStatus) throws Exception;
+    void check(EhrStatus ehrStatus);
 
 
     /**
--- a/api/src/main/java/org/ehrbase/api/service/ValidationService.java
+++ b/api/src/main/java/org/ehrbase/api/service/ValidationService.java
@@ -19,6 +19,7 @@
 package org.ehrbase.api.service;
 
 import com.nedap.archie.rm.composition.Composition;
+import com.nedap.archie.rm.ehr.EhrStatus;
 
 import java.util.UUID;
 
@@ -55,6 +56,14 @@ public interface ValidationService {
      * @throws IllegalArgumentException
      */
     void check(Composition composition) throws Exception;
+
+    /**
+     * initially check if ehrstatus is valid for further processing
+     * @param ehrStatus
+     * @throws IllegalArgumentException
+     */
+    void check(EhrStatus ehrStatus) throws Exception;
+
 
     /**
      * Remove all mappings in the cache
--- a/service/src/main/java/org/ehrbase/dao/access/jooq/StatusAccess.java
+++ b/service/src/main/java/org/ehrbase/dao/access/jooq/StatusAccess.java
@@ -41,6 +41,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.jooq.DSLContext;
 
+import java.sql.SQLException;
 import java.sql.Timestamp;
 import java.time.LocalDateTime;
 import java.util.List;
@@ -120,12 +121,21 @@ public class StatusAccess extends DataAccess implements I_StatusAccess {
     }
 
     public static I_StatusAccess retrieveInstanceByEhrId(I_DomainAccess domainAccess, UUID ehrId) {
-        StatusRecord record = domainAccess.getContext().fetchOne(STATUS, STATUS.EHR_ID.eq(ehrId));
+        try {
+            StatusRecord record = domainAccess.getContext().fetchOne(STATUS, STATUS.EHR_ID.eq(ehrId));
 
-        if (record == null)
-            return null;
+            if (record == null)
+                return null;
 
-        return createStatusAccessForRetrieval(domainAccess, record);
+            return createStatusAccessForRetrieval(domainAccess, record);
+        }
+        catch (Exception e){
+            if (e instanceof SQLException){
+                //retrieve the cause and use the cause
+                throw new IllegalArgumentException(e.getCause().getMessage());
+            }
+        }
+        return null;
     }
 
     /**
--- a/service/src/main/java/org/ehrbase/service/EhrServiceImp.java
+++ b/service/src/main/java/org/ehrbase/service/EhrServiceImp.java
@@ -206,6 +206,8 @@ public class EhrServiceImp extends BaseService implements EhrService {
                 throw (UnprocessableEntityException) e;
             if (e.getClass().equals(IllegalArgumentException.class))
                 throw new ValidationException(e);
+            if (e.getClass().equals(ValidationException.class))
+                throw e;
             else if (e.getClass().equals(org.ehrbase.validation.constraints.wrappers.ValidationException.class))
                 throw new ValidationException(e);
             else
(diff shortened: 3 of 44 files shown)


api/src/main/java/org/ehrbase/api/service/ValidationService.java
```diff
@@ -56,6 +57,14 @@
      */
     void check(Composition composition) throws Exception;
 
+    /**
+     * initially check if ehrstatus is valid for further processing
+     * @param ehrStatus
+     * @throws IllegalArgumentException
+     */
+    void check(EhrStatus ehrStatus);
```
jakesmolka, line 65: I removed the `throws Exception` here since it only throws `RuntimeException`s anyway. Alright? Would that make sense for the `check(Composition composition)`, too?

service/src/main/java/org/ehrbase/dao/access/jooq/StatusAccess.java
```diff
@@ -120,12 +121,21 @@ public static I_StatusAccess retrieveInstanceByParty(I_DomainAccess domainAccess
...
+            if (record == null)
+                return null;
 
-        return createStatusAccessForRetrieval(domainAccess, record);
+            return createStatusAccessForRetrieval(domainAccess, record);
+        }
+        catch (Exception e){
+            if (e instanceof SQLException){
+                //retrieve the cause and use the cause
+                throw new IllegalArgumentException(e.getCause().getMessage());
```
jakesmolka, line 135: Please use our own exceptions.

service/src/main/java/org/ehrbase/service/EhrServiceImp.java
```diff
@@ -58,22 +57,35 @@
...
+            validationService.check(status);
+        } catch (Exception e) {
+            // rethrow if this class, but wrap all others in InternalServerException
+            if (e.getClass().equals(UnprocessableEntityException.class))
+                throw (UnprocessableEntityException) e;
+            if (e.getClass().equals(IllegalArgumentException.class))
+                throw new ValidationException(e);
+            if (e.getClass().equals(ValidationException.class))
+                throw e;
+            else if (e.getClass().equals(org.ehrbase.validation.constraints.wrappers.ValidationException.class))
```
jakesmolka, line 83: We have two kinds of `ValidationException`, which is confusing. Do you know if we could replace the module specific `org.ehrbase.validation.constraints.wrappers.ValidationException` one with our project-wide `org.ehrbase.api.exception.ValidationException` one? I mean not just here, but generally?

service/src/main/java/org/ehrbase/service/ValidationServiceImp.java
```diff
@@ -131,6 +135,44 @@ public void check(Composition composition) throws Exception {
...
+        //first, check the built EhrStatus using the general Archie RM-Validator
+        List<RMObjectValidationMessage> rmObjectValidationMessages = RM_OBJECT_VALIDATOR.validate(ehrStatus);
+
+        if (!rmObjectValidationMessages.isEmpty()){
+            StringBuilder stringBuilder = new StringBuilder();
+            for (RMObjectValidationMessage rmObjectValidationMessage: rmObjectValidationMessages){
+                stringBuilder.append(rmObjectValidationMessage.toString());
+                stringBuilder.append("\n");
+            }
+            throw new ValidationException(stringBuilder.toString());
```
jakesmolka, line 154: I changed the exceptions to our own one. This is our general approach we decided on once. Would you agree here, or had you some custom use case in mind? If not, would you change the exception type in the other methods here, too?

service/src/main/java/org/ehrbase/service/ValidationServiceImp.java
```diff
@@ -131,6 +135,44 @@ public void check(Composition composition) throws Exception {
...
     }
 
+    @Override
+    public void check(EhrStatus ehrStatus) {
+
+        //case of a system generated ehr
+        if (ehrStatus == null)
+            return;
+
+        //first, check the built EhrStatus using the general Archie RM-Validator
```
jakesmolka, line 145: I changed the order of the checks. Now, first, the general RM validation is done and secondly, our more custom rules are applied. Alright?
The other way around I found it hard to differentiate between what I need to keep in mind and check with our `if`s VS. what the archie validator is going to check anyway. Now the broader and general check is applied first and probably catches all usual problems.

service/src/main/java/org/ehrbase/service/ValidationServiceImp.java
```diff
@@ -131,6 +135,44 @@ public void check(Composition composition) throws Exception {
...
+        if (ehrStatus.getSubject().getExternalRef() != null) {  // external_ref has 0..1 multiplicity, so null itself is okay
+            // but if it is there it has to have an ID
+            if (ehrStatus.getSubject().getExternalRef().getId() == null || ehrStatus.getSubject().getExternalRef().getId().getValue().isEmpty())
+                throw new ValidationException("ExternalRef ID is required");
+            // and a namespace
+            if (ehrStatus.getSubject().getExternalRef().getNamespace() == null) {
+                throw new ValidationException("ExternalRef namespace is required");
+                // which needs to be valid
+            } else if (!NAMESPACE_PATTERN.matcher(ehrStatus.getSubject().getExternalRef().getNamespace()).matches())
+                throw new ValidationException("Subject's namespace format invalid");
```
jakesmolka, line 171: I added the namespace check here, which was needed by an integration test. I reordered the conditionals here a bit.

tests/robot/EHR_SERVICE_TESTS/B.1_CREATE_EHR/B.1__a)_New_EHR.robot
```diff
@@ -342,7 +326,7 @@ MF-025 - Create new EHR w/ invalid subject (POST /ehr variants)
     [Documentation]     Covers edge case where subject is provided but is invalid
     ...                 because of some mandatory elements missing.
     [Tags]
-    [Template]          create ehr from data table
+    [Template]          create ehr from data table (invalid)
```
jakesmolka, line 329: @testautomation needs to do a quick sanity check on the changes of the test itself

service/src/main/java/org/ehrbase/service/ValidationServiceImp.java
```diff
@@ -131,6 +135,44 @@ public void check(Composition composition) throws Exception {
...
+            return;
+
+        //first, check the built EhrStatus using the general Archie RM-Validator
+        List<RMObjectValidationMessage> rmObjectValidationMessages = RM_OBJECT_VALIDATOR.validate(ehrStatus);
+
+        if (!rmObjectValidationMessages.isEmpty()){
+            StringBuilder stringBuilder = new StringBuilder();
+            for (RMObjectValidationMessage rmObjectValidationMessage: rmObjectValidationMessages){
+                stringBuilder.append(rmObjectValidationMessage.toString());
+                stringBuilder.append("\n");
```
jakesmolka, line 152: I observed a wrong input body (f.e. without `name`) returns the message from this string builder like intended, but the new lines are just literal `\n`, instead of a line break.

chevalleyc:
@jakesmolka thanks for the changes! I will merge this PR as to prevent it to block the next one (in particular PR #209 )

Issue #252, opened 2020-06-03
AQL REST: Cannot translate a null SQLException


ppazos:
Posted this query:

```
{
  "q": "SELECT c/uid/value FROM EHR e[ehr_id/value=$ehr_id] CONTAINS COMPOSITION c[openEHR-EHR-COMPOSITION.report-result.v1] CONTAINS OBSERVATION o[openEHR-EHR-OBSERVATION.lab_test-blood_glucose.v1] WHERE o/data[at0001]/events[at0002]/data[at0003]/items[at0078.2]/value/magnitude > 300",
  "query_parameters": {
    "ehr_id": "__set_by_testehr__"
  }
}
```

I got:

```
{
  "error": "Cannot translate a null SQLException".
  "status": "Bad Request"
}
```

From the error I'm not sure where the issue is.

This is the OPT:

[LabResults1.opt.txt](https://github.com/ehrbase/ehrbase/files/4724760/LabResults1.opt.txt)

If I remove the WHERE, the query is executed correctly.

chevalleyc:
Hi @ppazos could you please post a composition example? Thanks!

ppazos:
@chevalleyc of course,

This is another query that I get the same result:
```

{
	"q": "SELECT c/uid/value FROM EHR e[ehr_id/value=$ehr_id] CONTAINS COMPOSITION c[openEHR-EHR-COMPOSITION.referral.v1] CONTAINS INSTRUCTION i[openEHR-EHR-INSTRUCTION.request-referral.v1] CONTAINS CLUSTER cl[openEHR-EHR-CLUSTER.individual_personal.v1] WHERE cl/items[at0017]/value/defining_code/code_string = 'at0020'",
	"offset": 0,
	"fetch": 20,
	"query_parameters": {
		"ehr_id": "e5fa5f0e-07d3-4c67-8f1a-78fb7a7a0dd6"
	}
}
```

And this is one of the COMPOs committed using the OPT associated with that query ^ (the commit was successful 204)

```
{
	"_type": "COMPOSITION",
	"name": {
		"_type": "DV_TEXT",
		"value": "Referral document"
	},
	"archetype_details": {
		"archetype_id": {
			"value": "openEHR-EHR-COMPOSITION.referral.v1"
		},
		"template_id": {
			"value": "Referral"
		},
		"rm_version": "1.0.2"
	},
	"archetype_node_id": "openEHR-EHR-COMPOSITION.referral.v1",
	"language": {
[...]

PR #253, opened 2020-06-05
Fix issue when implicit where ehr/id/value is passed from EHR predicate

chevalleyc:
## Changes

This fixes the issue whenever the ehr_id/value is specified in the from EHR predicate.

## Related issue

Fixes: #252 


ppazos:
@chevalleyc can this be merged?

PR #268, opened 2020-07-02
Feature/291 update aql terminology grammar with sec

lmarcoruiz:
## Changes

Adapts parser to the latest openEHR SEC agreements.
Implements Validate operation with external terminology server.

## Related issue

Closes https://github.com/ehrbase/project_management/issues/291

### DIFF
--- a/service/src/main/java/org/ehrbase/service/FhirTerminologyServerR4AdaptorImpl.java
+++ b/service/src/main/java/org/ehrbase/service/FhirTerminologyServerR4AdaptorImpl.java
@@ -171,7 +171,7 @@ public final class FhirTerminologyServerR4AdaptorImpl
 						String.class);
 				String response = responseEntity.getBody();
 				DocumentContext jsonContext = JsonPath.parse(response);
-				Boolean result = (Boolean) ((JSONArray)jsonContext.read("$.parameter[:1].valueBoolean")).get(0);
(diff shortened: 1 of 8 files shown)


service/src/main/java/org/ehrbase/service/FhirTerminologyServerR4AdaptorImpl.java
```diff
@@ -147,6 +147,7 @@ public final DvCodedText lookUp(final String conceptId) {
 	@Override
 	public final Boolean validate(final DvCodedText concept, final String valueSetId) {
 		// TODO Auto-generated method stub
+		System.out.println("inside the validate method of R4 implemetntation");
```
chevalleyc, line 150: I would suggest to use a logger and use a log.debug(...) instead

Geisterfalle: As Luis is absent for some time, could you please do the changed and merge?

chevalleyc: Done

jakesmolka:
FYI: @lmarcoruiz @chevalleyc those changes to the service pom.xml you just merged are unclean (different formatting) and are creating merge conflicts, see this direct link:
https://github.com/ehrbase/ehrbase/pull/268/files#diff-15f45f0e05aff8eda200826913a1367e

I'll try to revert it, while keeping possible real changes via piggybacking on #267

chevalleyc:
I didn't look at the pom.xml TBH, just integrated the change I suggested. Sorry about that

PR #281, opened 2020-07-24
Add duplicate detection for directory

axelsiebert:
## Changes

- Add detection for already existing Directory of an EHR
- Respond with 409 - CONFLICT in case of detected folder

## Related issue

Fixes https://github.com/ehrbase/project_management/issues/231

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Detect duplicates on POST Directory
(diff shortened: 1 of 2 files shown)


axelsiebert:
> @axelsiebert looks good. But seems we have no test that covers that case. We have a test that checks "validate POST response - 409 folder already exists" which is failing silently because it's marked as not-ready (it's related to [ehrbase/project_management#148](https://github.com/ehrbase/project_management/issues/148))
> 
> I'm not sure if we need a new tests or have to refactor the existing one. We should should create a follow up ticket.

Sorry I made a mistake with the description of changes. The response code will be 409 as you mentioned. Then the test could be marked as ready.

Issue #284, opened 2020-08-03
Timezone not working


stefanspiska:
For the example files see timezone.zip

When I send timezone_send.json
which has
context.start_time.value= 2020-05-03T14:35:16,583Z
context.end_time.value=2020-08-03T14:35:16,583+02:00

and then make a get for the saved composition
then I get:
context.start_time.value= 2020-05-03T16:35:16,583Z
context.end_time.value=2020-08-03T14:35:16,583+02:00

so the time which was utc has moved exactly two hours. 

My local timezone is "Europe/Berlin"

stefanspiska:
[timezone.zip](https://github.com/ehrbase/ehrbase/files/5016076/timezone.zip)

stefanspiska:
ps. with commit https://github.com/ehrbase/ehrbase/commit/f3c9bba5f9be5f4e9dd1691b0a999ce90784d2ee was it still working

PR #286, opened 2020-08-06
fixed time drift related to local timezone

chevalleyc:
## Changes

Present a date/time related to the timezone that was given in the DvDateTime variable.

NB. in postgres, date/time are always stored as an SQL timestamp relatively to the local (system) timezone. The logic is then to keep this conversion and calculate the DvDateTime according to the saved timezone.

## Related issue

Closes https://github.com/ehrbase/ehrbase/issues/284


chevalleyc:
Fixed the change request. Issue is that DB date/time is according to the locale. To translate it to another locale, this requires:
1. Adjust the date/time to UTC
2. Apply the timezone
3. Adjust the result to UTC
4. Format the result ISO8601 with the stored timezone.

NB. The timezone is an offset (f.e. +02:00), literal (f.e. 'CET') is not supported by Archie

For more on TimeZone and PostgreSQL: https://www.enterprisedb.com/postgres-tutorials/postgres-time-zone-explained

PR #311, opened 2020-09-13
Automatic Docker image releases via DockerHub's Autobuilds

wlad:
An EHRbase docker image is created and published on DockerHub regestry:
- Pushes/merges to ~~develop~~ master branch publish a docker image w/ `latest` tag, i.e. **ehrbaseorg/ehrbase:latest**
- Pushes/merges to develop branch publish a docker image w/ `next` tag, i.e. **ehrbaseorg/ehrbase:next**
- Pushes to release-* branches publish a docker image w/ version `1.2.3` tag, i.e. **ehrbaseorg/ehrbase:0.13.0**
- Uses DockerHub infrastructure (no resources of CircleCI used)

## Changes

- Adds multi stage Dockerfile with scripts to build ehrbase inside a docker container

## Related issue

- closes https://github.com/ehrbase/project_management/issues/333
- closes https://github.com/ehrbase/project_management/issues/342

## DockerHub Configuration Docs
It was not sufficient to connect our ehrbase-tech-user, instead the "real" owner of the Github Org had to be connected to enable Autobuild on Docker Hub. Now that the connection is established everything else can be configured w/ the ehrbase-tech-user login to DockerHub (docker.com)

image
---
image
---
image

### DIFF
--- a/.docker_scripts/create-ehrbase-user.sh
+++ b/.docker_scripts/create-ehrbase-user.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+set -e
+
+psql --username "$POSTGRES_USER" --dbname "$POSTGRES_DB" <<-EOSQL
+    CREATE ROLE ${EHRBASE_USER} LOGIN PASSWORD '${EHRBASE_PASSWORD}';
+    CREATE DATABASE ehrbase ENCODING 'UTF-8' TEMPLATE template0;
+    GRANT ALL PRIVILEGES ON DATABASE ehrbase TO ${EHRBASE_USER};
+    CREATE USER root WITH SUPERUSER;
+EOSQL
+
+# Stop database before proceeding
+su - postgres -c "pg_ctl -D ${PGDATA} -w stop"
--- a/.docker_scripts/docker-entrypoint.sh
+++ b/.docker_scripts/docker-entrypoint.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+echo 
+echo "EHRBASE_VERSION: $(cat ehrbase_version)"
+java -Dspring.profiles.active=docker -jar ehrbase.jar
--- a/.docker_scripts/install-jsquery.sh
+++ b/.docker_scripts/install-jsquery.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+set -e
+
+# Fetch from remote repository
+git clone https://github.com/postgrespro/jsquery.git
+cd jsquery
+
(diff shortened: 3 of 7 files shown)


wlad:
## ⚠️  displayed fail is just a Github/Sonarcloud UI glitch!
image

## everything is fine
image

jakesmolka:
Just to clarify: All other image tags except `latest` and `$version` can be deleted and won't be necessary anymore in the near future? That would be very good!

axelsiebert:
I think it is a convention to use the "latest" tag for an alias of the last release version, e.g. for version 0.13.0 the latest also points to that version. If there are other versions as 0.12.0 the are available but "latest" does not point to that.

For develop I would suggest to add a tag like "next" or "dev" that indicates it is a testing and maybe unstable version.

wlad:
@jakesmolka @axelsiebert 

> For develop I would suggest to add a tag like "next" or "dev" that indicates it is a testing and maybe unstable version.

Since there is no standard for it I think it's a matter of what is stated in our README so that anyone who wants to use it knows what to expect from given tags.

We can do something like below w/o much effort (is just little configuration in DockerHub UI):

master --> ehrbaseorg/ehrbase:**latest**     (where latest == 0.13.0 as per this example)
realease-* --> ehrbaseorg/ehrbase:{VERSION} i.e. ehrbaseorg/ehrbase:**0.13.0**
develop --> ehrbaseorg/ehrbase:**next** OR ehrbaseorg/ehrbase:**dev**

A possible issue with this approach may be that latest may end up being outdated most of the time - because the version is not updated very often (yet). That's why I am actually for having only latest and version tags (but: only if ehrbase/project_management#364 turns out positive - this would ensure that an image is only published if it passed all tests - otherwise latest/version/devORnext approach will be more secure)

wlad:
@axelsiebert  I've updated the rules on Docker Hub

image

the last one will create something like `ehrbaseorg/ehrbase:0.13.0` when commit is pushed to a branch `release-0.13.0`

wlad:
@axelsiebert  I've updated Sphinx docs --> https://github.com/ehrbase/documentation/pull/12

axelsiebert (approved):
Changes are fine. I like the clean approach of splitting the scripts into separate files.

For the docker tagging I think there is some additional work with more effort required to make it clean. As the Docker Hub UI only provides basic settings we should add a CircleCI script that does the deployment to docker based on dates.

**Note**: The release branches contain always a not released version. The version number tag should be fetched from tag in master branch.

PR #312, opened 2020-09-14
Feature/pre load caches

stefanspiska:
## Changes

-  Precalculate containment tree from OPT template


stefanspiska:
changes will be merged together with https://github.com/ehrbase/ehrbase/tree/feature/ehrbase/project_management/issues/367_new_opt_parser

PR #314, opened 2020-09-18
Feature/ehrbase/project management/issues/367 new opt parser

stefanspiska:


jakesmolka (approved):
Failing tests are due to EHRbase and SDK are changed simultaneously. This will be resolved manually.

PR #344, opened 2020-10-14
Admin API endpoints (Feature/314 admin delete ehr)

jakesmolka:
## Changes

Adds Admin API endpoints: Del EHR, Del Composition and Del Contribution. All are cascading into all linked entities for their deletion as well.

See the documentation page in Jira for implementation details.

## Related issue

closes https://github.com/ehrbase/project_management/issues/314
closes https://github.com/ehrbase/project_management/issues/320
closes https://github.com/ehrbase/project_management/issues/389


chevalleyc:
Could you please give the pointer to the doc ?

chevalleyc:
1. I think deleting a party_identified and system could be performed with a simple function as below:

```
CREATE OR REPLACE FUNCTION delete_party_identified(party_id UUID)
  RETURNS boolean AS 
$func$
BEGIN
   DELETE FROM ehr.party_identified WHERE party_identified.id = party_id; -- or any other identification

   IF NOT FOUND THEN
      RETURN NULL;                        -- UUID not found, return NULL
   END IF;

   RAISE SQLSTATE 'DELPI';                -- If DELETE, raise custom exception (5 positions literal!)

   EXCEPTION
   WHEN FOREIGN_KEY_VIOLATION THEN
      RETURN FALSE;
   WHEN SQLSTATE 'DELPI' THEN
   -- logging should be added here to record: who, when, deleted_object id
      RETURN TRUE;
END  
$func$ LANGUAGE plpgsql;
```

that said, only the party_identifed belonging to the EHR should be delete IMO (that is the subject, the owner of the EHR). The other party (f.e. participant, composer) may be provided by another system upon registration. I would suggest to have a 'clean-up' admin endpoint to allow this specific operation.

The same applied to system.

2. The deletion should be logged (https://www.postgresql.org/docs/current/plpgsql-errors-and-messages.html)

jakesmolka:
@wlad Does the sonarcloud code analysis checks the whole project or only the PR's changes here? The "1 Bug" is about a line I never touched in this PR (see what it points to vs. what I did in that file https://github.com/ehrbase/ehrbase/pull/344/files#diff-923a7421d7cc516ab7e680ff6aefc35fe80cefc4d639218a53375e6e8b64f201).
Also the analysis was successful and then suddenly assumes those changes here? I'm confused :D

jakesmolka:
Regarding https://github.com/ehrbase/project_management/issues/318:

> A simple fix like
> 
> -- fix bug 318: directory foreign key in ehr table
> ALTER TABLE ehr.ehr
>     DROP COLUMN directory,
>     ADD COLUMN directory UUID references ehr.folder(id);
> 
> will break Folder/Directory PUT or updating. So that would need to be refactored to fix this issue.

So this is definitely out of scope of this feature. Therefore, the malfunctioning fix has been removed again.

Issue #324, opened 2020-10-02
AQL 'Select e From EHR e' throws nullpointer


yampeku:
## Steps to reproduce ## 

Launch ad-hoc query 

`"q":"Select e From EHR e"`

## Actual result ##

Throws nullpointer exception, seems to be searching the query in the stored queries
```
{
  "error": "Could not retrieve stored query, reason: java.lang.NullPointerException",
  "status": "Bad Request"
}
```

## Expected result ##

Return all EHR available in the repo, or '400 bad request' if query is unsupported

## Success criteria ##

- [ ] The defect has been fixed
- [ ] The defect is checked by an unit or an integration test (Robot)
- [ ] Merge Request approved
- [ ] Unit tests passed
- [ ] Build without errors
- [ ] Release notes prepared
- [ ] Runtime warnings

yampeku:
Doing more testing seems to be related with the "Select e" part of the query

PR #348, opened 2020-10-26
Feature/324 select e

chevalleyc:
## Changes

support AQL query as `SELECT e FROM EHR e`

Few limitations related to filtering (e.g. WHERE clause):

- where on ARRAY is not yet supported (see https://github.com/ehrbase/project_management/issues/404). Suggest to retrieve the whole set and perform opt out as required for the time being.
- further, WHERE using complex expression (that is implicit SQL functions) is not yet supported. This may result in confusing error messages. See https://github.com/ehrbase/project_management/issues/407 for more on this.

## Related issue

Fixes: https://github.com/ehrbase/ehrbase/issues/324


axelsiebert (approved):
Folder related changes are fine for me. Approve the PR for this part. 

Waiting with merge for review from @birgerhaarbrandt

PR #393, opened 2020-11-27
Status endpoint for version check

axelsiebert:
## Changes

- Add endpoint controller to serve from http://${ehrbase_host:port}/ehrbase/rest/openehr/v1/status
- Add DatabaseStatusAccess to retrieve the PostgreSQL server version from running database
- Add stored procedure to migrations to get the full version string from database
- Add service to extract OS, JVM, PostgreSQL and dependency versions from running instance
- Add build property exposure to build process with maven

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Adds Admin API endpoints: Del EHR, Del Composition and Del Contribution (see: https://github.com/ehrbase/ehrbase/pull/344)
 - Add ATNA logging configuration capabilities (see https://github.com/ehrbase/ehrbase/pull/355)
 - Support for EHR_STATUS and (partial) FOLDER version objects in contributions (see: https://github.com/ehrbase/ehrbase/pull/372)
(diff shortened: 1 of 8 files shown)


chevalleyc (commented):
Is there a way to get the build time, build author and GIT commit hash for ehrbase, sdk and archie. The issue I see is that the build details from the pom are not reflecting the actual code base version unless we strictly enforce version numbering (but how?).

axelsiebert:
> Is there a way to get the build time, build author and GIT commit hash for ehrbase, sdk and archie. The issue I see is that the build details from the pom are not reflecting the actual code base version unless we strictly enforce version numbering (but how?).

The Archie and openEHR_SDK commit hashes are currently extracted from pom.xml. We can use these information to build and debug the EHRbase with the same dependencies. I think there is no way to build the EHRbase with a local "non-commited" state of openEHR_SDK or Archie. And even if this is possible, a bug issue raised from this setup cannot be accepted due to not reproducable (imo.!)

The build time of EHRbase itself should not be a problem. I will integrate this later.

I think the build author does not provide valuable information and introduce the danger to point to someone from the team to be responsible for a bug. 

Please correct me if I misunderstood you ;-)

PR #407, opened 2021-01-08
Feature/404 support set in where

chevalleyc:
## Changes

NB. the UC tests are located in package org.ehrbase.aql.sql.queryImpl.translator.testcase (module service)

- Support set returning functions in WHERE clause (see tests UC27 - UC33)
- refactored SQL translation on JSON structure to better support arrays, ease maintenance and allow future json querying or dialect (in particular JSONPATH)
- Atomized tests (see above) to ease maintenance and document current expression support, enable different tests approach depending on selected dialect (f.e. jsquery or plain SQL)
- Enhanced AQL grammar to support or fixed further set operators (IN, ANY, SOME, ALL, BETWEEN)
- added some SQL primitives: NULL, UNKNOWN, TRUE, FALSE
- fixed some last minutes translation errors (unresolved variables)
- enable functions: COUNT, MIN, MAX, AVG

## Related issue

fixes: https://github.com/ehrbase/project_management/issues/404
fixes: https://github.com/ehrbase/project_management/issues/442


birgerhaarbrandt:
@chevalleyc I found that MIN( ) does not work on datetimes and gives back error messages from postgres. I think this requires a second look

PR #413, opened 2021-01-25
fixed wrong datatype cast for ehr/time_created/value

chevalleyc:
## Changes

fix `ehr/time_created/value` to render as VARCHAR instead of JSONB.

This was causing the HTTP body encoding to be truncated as the JSONB field couldn't be interpreted... as a result only the columns before this bad representation were displayed.

## Related issue

related to ongoing tests on https://github.com/ehrbase/project_management/issues/205


wlad (approved):
looks good :+1:

PR #419, opened 2021-01-29
Feature/206 unquoted substitution

chevalleyc:
## Changes

Fix variable substitution in node archetype id predicate by handling the quoted argument properly (remove the single quotes). 
This fix uses parse error recovery at ANTLR parsing time. Hence, the fix is compatible with future grammar enhancement.

## Related issue

fixes: https://github.com/ehrbase/project_management/issues/206


wlad (approved):
http 400 bad request error does not occur any more

I removed bug traces form related tests

PR #424, opened 2021-02-03
handle unresolved contains properly

chevalleyc:
## Changes

Handle properly the columns in resultset whenever contains is not resolved.

## Related issue

Fixes: https://github.com/ehrbase/project_management/issues/462


birgerhaarbrandt:
@chevalleyc Would it be helpful to have this scenario also covered by additional integration tests inside the SDK?

chevalleyc:
> @chevalleyc Would it be helpful to have this scenario also covered by additional integration tests inside the SDK?

Indeed, but it is quite tricky to reproduce since we need to deal with "emptiness" (empty DB, unrelated templates etc.)

birgerhaarbrandt:
@chevalleyc looks all good to me but consider the cover this scenario in the integration tests.

PR #456, opened 2021-03-03
ensure json object is not iterated

chevalleyc:
## Changes

Fix an issue with json array iteration as it could fall on an non-iterable object.

## Related issue

Fixes: https://github.com/ehrbase/project_management/issues/506 (see comment in the issue re particular failing select item)

### DIFF
--- a/service/src/main/java/org/ehrbase/aql/sql/queryimpl/IterativeNode.java
+++ b/service/src/main/java/org/ehrbase/aql/sql/queryimpl/IterativeNode.java
@@ -22,11 +22,7 @@ import org.ehrbase.dao.access.interfaces.I_DomainAccess;
 import org.ehrbase.ehr.util.LocatableHelper;
 import org.ehrbase.service.IntrospectService;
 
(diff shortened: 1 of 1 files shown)


birgerhaarbrandt:
@chevalleyc Is this covered by an integration test in the SDK?

chevalleyc:
> @chevalleyc Is this covered by an integration test in the SDK?

not yet... it is based on Pascal's Viro test suite. This one should be in the SDK test, but not there yet.

birgerhaarbrandt:
@chevalleyc I would rather like to have this as part of the PR or in a ticket that is directly addressed next. What do you prefer?

chevalleyc:
> @chevalleyc I would rather like to have this as part of the PR or in a ticket that is directly addressed next. What do you prefer?

I much prefer a separate ticket as editing tests is not practical at the moment and still cause issue with the sync feature

birgerhaarbrandt:
I will create a separate ticket then

PR #499, opened 2021-05-05
Attribute based access controll support

jakesmolka:
## Changes

- Adds ABAC integration support
- Adds integration tests (needs fresh DB and are disabled by default)

A lot of configuration possible. See `application.yml` for information.

For a meta-perspective see `AbacIntegrationTest.java`.

For the implementation itself the `CustomMethodSecurityExpressionRoot`.

The whole feature is rather huge and complex, so see https://wiki.vitagroup.ag/display/ETHERCIS/ABAC+Integration for a better overview.

## Related issue

closes https://github.com/ehrbase/project_management/issues/505

- [x] Create separate issue to add the ABAC tests to the CI pipeline
- [x] Push documentation to public readthedocs documentation (the documentation will just be added into the current structure, see https://github.com/ehrbase/project_management/issues/521)


application/src/main/java/org/ehrbase/application/abac/AbacCheck.java
```diff
@@ -0,0 +1,79 @@
...
+        .writerWithDefaultPrettyPrinter()
+        .writeValueAsString(bodyMap);
+
+    HttpRequest request = HttpRequest.newBuilder()
+        .uri(URI.create(url))
+        .header("Content-Type", "application/json")
+        .POST(BodyPublishers.ofString(requestBody))
+        .build();
+
+    return HttpClient.newHttpClient().send(request, BodyHandlers.ofString());
```
subigre, line 73: It is not directly linked to this task but as a potential (future?) improvement, I would suggest to that we could reuse the HttpClient instance configured through `ClientConfiguration` and `ClientProperties`.
I do not know i EHRbase requires other external accesses than ABAC server and remote terminology servers?
It could be nice to have common configurations for all HTTP clients. For instance, if EHRbase running behind an authenticated  proxy or if an ABAC instance also requires two-way authentication.

application/src/main/java/org/ehrbase/application/abac/AbacConfig.java
```diff
@@ -0,0 +1,164 @@
...
+@Configuration
+@EnableConfigurationProperties
+@ConfigurationProperties(prefix = "abac")
+public class AbacConfig {
+
+  private URI server;
+  private String organizationClaim;
+  private String patientClaim;
+  // ehrStatus
+  @Value("${abac.policy.ehrstatus.name}")
```
subigre, line 36: I would go either for `@ConfigurationProperties` either for `@Value` but having both looks strange to me.

application/src/main/java/org/ehrbase/application/config/SecurityConfig.java
```diff
@@ -30,6 +30,7 @@
 import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
```
subigre, line 30: For the readability and maintenance, I think that we should create separate configuration classes like : `SecurityConfiguration`, `BasicSecurityConfigucation`, `OAuth2SecurityConfiguration `
And we could enable the correct one using `@ConditionalOnProperty `annotation

application/src/main/java/org/ehrbase/application/abac/CustomMethodSecurityExpressionRoot.java
```diff
@@ -0,0 +1,523 @@
...
+        policyParameters = new ArrayList<>(Arrays.asList(abacConfig.getPolicyContributionParameters()));
+        requestUrl = requestUrl.concat(abacConfig.getPolicyContributionName());
+        break;
+      case BaseController.QUERY:
+        policyParameters = new ArrayList<>(Arrays.asList(abacConfig.getPolicyQueryParameters()));
+        requestUrl = requestUrl.concat(abacConfig.getPolicyQueryName());
+        break;
+      default:
+        throw new InternalServerException("ABAC: Invalid type given from Pre- or PostAuthorize");
+    }
```
subigre, line 176: Linked to my comment on `AbacConfig`, I think that using `@ConfigurationProperties`  with a Map<String, PolicyParameters> could be simpler here.

rest-openehr/src/main/java/org/ehrbase/rest/openehr/controller/OpenehrCompositionController.java
```diff
@@ -84,11 +83,13 @@
 
     @Autowired
     public OpenehrCompositionController(CompositionService compositionService) {
-
         this.compositionService = Objects.requireNonNull(compositionService);
     }
 
     @PostMapping(value = "/{ehr_id}/composition", consumes = {"application/xml", "application/json"})
+    // checkAbacPre /-Post attributes (type, auth object, subject, payload, content type)
+    @PreAuthorize("checkAbacPre(@openehrCompositionController.COMPOSITION, authentication, "
```
subigre, line 91: This could be simplified based on the comment in`CustomMethodSecurityExpressionRoot`

rest-openehr/src/main/java/org/ehrbase/rest/openehr/controller/OpenehrVersionedEhrStatusController.java
```diff
@@ -106,6 +107,9 @@ public OpenehrVersionedEhrStatusController(EhrService ehrService, ContributionSe
     }
 
     @GetMapping(path = "/version")
+    // checkAbacPre /-Post attributes (type, auth object, subject, payload, content type)
+    @PreAuthorize("checkAbacPre(@openehrVersionedEhrStatusController.EHR_STATUS, authentication, "
```
subigre, line 111: This could be simplified based on the comment in `CustomMethodSecurityExpressionRoot`

rest-openehr/src/main/java/org/ehrbase/rest/openehr/controller/OpenehrCompositionController.java
```diff
@@ -315,6 +322,9 @@ public ResponseEntity deleteComposition(@ApiParam(value = REQ_OPENEHR_VERSION) @
      * "{?version_at_time}" is hidden in swagger-ui, it only is here to be piped through.
      */
     @GetMapping("/{ehr_id}/composition/{version_uid}")
+    // checkAbacPre /-Post attributes (type, auth object, subject, payload, content type)
+    @PostAuthorize("checkAbacPost(@openehrCompositionController.COMPOSITION, authentication, "
+        + "@ehrService.getSubjectExtRef(#ehrIdString), returnObject, #accept)")
```
subigre, line 327: I will test now because I am not 100% sure may be the `returnObject` is not needed here

application/src/main/java/org/ehrbase/application/abac/CustomMethodSecurityExpressionRoot.java
```diff
@@ -0,0 +1,523 @@
...
+   * @param type    Type of scope's resource
+   * @param auth    Authentication stored in the security context
+   * @param subject Subject ID from the current EHR context
+   * @param payload Payload object, either request's input or response's output
+   * @param contentType Content type from the scope
+   * @return True if ABAC authorizes given attributes
+   * @throws IOException On parsing error
+   * @throws InterruptedException On error while communicating with the ABAC server
+   */
+  public boolean checkAbacPost(String type, Authentication auth, String subject, Object payload,
```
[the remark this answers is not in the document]
jakesmolka: Good point! I removed the `auth` parameter from the signature and used `this.authentication` in the method instead.

application/src/main/java/org/ehrbase/application/abac/MethodSecurityConfig.java
```diff
@@ -0,0 +1,65 @@
...
+  private final AbacConfig abacConfig;
+
+  public MethodSecurityConfig(AbacConfig abacConfig) {
+    this.abacConfig = abacConfig;
+  }
+
+  @Value("${abac.disabled}")
+  private boolean abacDisabled;
+
+  /**
```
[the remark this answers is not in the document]
jakesmolka: Also a good idea! I think I managed to refactor it successfully.

application/src/main/java/org/ehrbase/application/config/SecurityConfig.java
jakesmolka: Good idea! I refactored the config classes to be separated like you suggest. Can you check again if that looks okay?

application/src/main/java/org/ehrbase/application/abac/CustomMethodSecurityExpressionRoot.java
jakesmolka: I couldn't get rid of this switch structure here, but the `AbacConfig` is much leaner with its own static class und enums, so the setting are type safe. Thanks!

application/src/main/java/org/ehrbase/application/abac/AbacCheck.java
jakesmolka: I added the foundation for this in the last commit. I does only allow to set a proxy yet, because I'm not sure how real auth requirements would look like. Can you recheck this too, and check if this is the direction you had in mind?

subigre (approved):
Looks good but I added two comments

application/src/main/java/org/ehrbase/application/config/HttpClientConfig.java
```diff
@@ -0,0 +1,80 @@
...
+import java.net.http.HttpClient.Version;
+import java.time.Duration;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@EnableConfigurationProperties
+@ConfigurationProperties(prefix = "httpclient")
+public class HttpClientConfig {
```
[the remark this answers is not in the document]
jakesmolka: Oh interesting, didn't knew it was there. I'm not sure now. Why is yours in the `service` package? In the end two different http client config are obviously not a good practice. But it feels wrong to pull that client from the `service` layer into my ABAC stuff.

application/src/main/java/org/ehrbase/application/abac/AbacCheck.java
```diff
@@ -0,0 +1,88 @@
...
+import java.net.http.HttpRequest.BodyPublishers;
+import java.net.http.HttpResponse;
+import java.net.http.HttpResponse.BodyHandlers;
+import java.util.Map;
+import org.ehrbase.api.exception.InternalServerException;
+import org.ehrbase.application.config.HttpClientConfig;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.stereotype.Component;
+
+@ConditionalOnProperty(name = "abac.enabled")
```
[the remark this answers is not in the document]
jakesmolka: Thanks! I refactored it.

application/src/main/java/org/ehrbase/application/config/HttpClientConfig.java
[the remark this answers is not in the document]
jakesmolka: Since there was no activity for 7 days I extracted that into the following issue, so I can continue with this PR here.
https://github.com/ehrbase/project_management/issues/522

Issue #498, opened 2021-04-29
AQL query returning null when different nesting level are present in the composition


svituz:
## Configuration information

- EHRbase version: 0.16.5
- openEHR_SDK version: bcc19c2
- Archie version: 5a4a9442e5e2516aed5f25636ac236682aace7b9
- PostgreSQL version: PostgreSQL 11.5 on x86_64-pc-linux-musl, compiled by gcc (Alpine 8.3.0) 8.3.0, 64-bit
- Java Runtime version: Oracle Corporation 11.0.10+9
- Operating System version: "Linux amd64 4.15.0-45-generic",

## Description

If a template and a related composition presents two sections with the same child type (e.g., an ACTION) but the child is at different nesting level, when querying the composition for the child, it returns null

## Steps to reproduce ## 

- Download the template from [here](https://gist.github.com/svituz/e2569bb9a870fb55951fdde0d44728c9#file-openehr_example_template) and create it in ehrbase using the REST API
- create an EHR
- Download the composition from [here](https://gist.github.com/svituz/e2569bb9a870fb55951fdde0d44728c9#file-ehrbase_aql_bug_null_result-json) and create it in ehrbase using the REST API
- Perform the query 
```
select e/ehr_id/value, a/description[at0001]/items[at0002]/name/value from EHR e contains ACTION a
```

## Actual result ##

```
{
    "q": "select e/ehr_id/value, a/description[at0001]/items[at0002]/name/value from EHR e contains ACTION a",
    "columns": [
        {
            "path": "/ehr_id/value",
            "name": "#0"
        },
        {
            "path": "/description[at0001]/items[at0002]/name/value",
            "name": "#1"
        }
    ],
    "rows": [
        [
            "e1f54616-338a-4eb6-86e0-621c248f339c",
            null
        ]
    ]
}
```
## Expected result (Acceptance Criteria) ##

```
{
    "q": "select e/ehr_id/value, a/description[at0001]/items[at0002]/name/value from EHR e contains ACTION a",
    "columns": [
        {
            "path": "/ehr_id/value",
            "name": "#0"
        },
        {
            "path": "/description[at0001]/items[at0002]/name/value",
            "name": "#1"
        }
    ],
    "rows": [
        [
            "e1f54616-338a-4eb6-86e0-621c248f339c",
            "Therapy"
        ]
    ]
}
```

## Definition of Done ##

- [ ] The defect is checked by an unit or an integration test (Robot)
- [ ] Merge Request approved
[...]

Issue #496, opened 2021-04-28
AQL queries not returning expected ACTIONS


svituz:
## Configuration information

- EHRbase version: 0.16.5
- openEHR_SDK version: bcc19c2
- Archie version: 5a4a9442e5e2516aed5f25636ac236682aace7b9
- PostgreSQL version: PostgreSQL 11.5 on x86_64-pc-linux-musl, compiled by gcc (Alpine 8.3.0) 8.3.0, 64-bit
- Java Runtime version: Oracle Corporation 11.0.10+9
- Operating System version: "Linux amd64 4.15.0-45-generic",
## Description

When using a template that has two SECTIONS and every section has a nested ACTION, querying for the ACTION returns only the child of the first SECTION and not both. 

## Steps to reproduce ## 

- Download the template from [here](https://gist.github.com/svituz/e2569bb9a870fb55951fdde0d44728c9#file-openehr_example_template-xml) and create it in ehrbase using the REST API
- create an EHR
- Download the composition from [here](https://gist.github.com/svituz/e2569bb9a870fb55951fdde0d44728c9#file-ehrbase_aql_bug_missing_result-json) and create it in ehrbase using the REST API
- Perform the query 
```
select e/ehr_id/value, a/description[at0001]/items[at0002]/name/value from EHR e contains ACTION a
```

## Actual result ##

The output is 
```
{
    "q": "select e/ehr_id/value, a/description[at0001]/items[at0002]/name/value from EHR e contains ACTION a",
    "columns": [
        {
            "path": "/ehr_id/value",
            "name": "#0"
        },
        {
            "path": "/description[at0001]/items[at0002]/name/value",
            "name": "#1"
        }
    ],
    "rows": [
        [
            "e1f54616-338a-4eb6-86e0-621c248f339c",
            "Surgery type"
        ]
    ]
}
```

## Expected result (Acceptance Criteria) ##

```
{
    "q": "select e/ehr_id/value, a/description[at0001]/items[at0002]/name/value from EHR e contains ACTION a",
    "columns": [
        {
            "path": "/ehr_id/value",
            "name": "#0"
        },
        {
            "path": "/description[at0001]/items[at0002]/name/value",
            "name": "#1"
        }
    ],
    "rows": [
        [
            "e1f54616-338a-4eb6-86e0-621c248f339c",
            "Surgery type"
        ],
        [
            "e1f54616-338a-4eb6-86e0-621c248f339c",
            "Therapy"
        ]
    ]
}
[...]

Issue #464, opened 2021-03-22
AQL path not unique for template nodes while creating WebTemplate


sasurfer:
## Configuration information

- EHRbase version:
- openEHR_SDK version:
- Archie version:
- PostgreSQL version:
- Java Runtime version:
- Operating System version:
<?xml version='1.0' encoding='UTF-8'?><status><ehrbase_version>0.15.0</ehrbase_version><openehr_sdk_version>fc328d5</openehr_sdk_version><archie_version>dc09642</archie_version><jvm_version>Oracle Corporation 11.0.10+9</jvm_version><os_version>Linux amd64 4.15.0-136-generic</os_version><postgres_version>PostgreSQL 11.5 on x86_64-pc-linux-musl, compiled by gcc (Alpine 8.3.0) 8.3.0, 64-bit</postgres_version></status>
## Steps to reproduce ## 

When I create a template with nested archetypes I receive a warning about aql paths not uniques while they are.

A simple template and the message displayed in EHRbase can be found at this address:

https://gist.github.com/sasurfer/509e778d6e6ecf603dd781e18cc4addb

## Definition of Done ##

- [ ] The defect is checked by an unit or an integration test (Robot)
- [ ] Merge Request approved
- [ ] Unit tests passed
- [ ] Build without errors
- [ ] Release notes prepared
- [ ] No additional runtime warnings

sasurfer:
The problem persist in the 0.16 version. The warning keep showing and an AQL search on that template doesn't give all the leaf values. In the example I posted only one of the two boolean values is returned. The other is returned as null.

chevalleyc:
Hi @sasurfer the reason of this warning is that we have decided to resolve containment clause on a meta-level using the webtemplate representation. In the past we used path indexing on actual compositions and this has proven to be a massive bottleneck in querying performance. 

Hence, whenever for a CONTAINS clause to be resolved it is expected that the path to an item stucture node is unique. 

As per specification `A node predicate defines criteria on fine-grained data.` (AQL), we have applied this reasoning to our resolution strategy. The benefit is that performance wise, querying is fast and stable regardless of the DB size (at least on the CONTAINS clause level obviously). I would recommend refactoring your template to suit this approach. Thanks!

sasurfer:
Hi @chevalleyc . I'm not sure I got your point. Are you saying that this kind of template cannot be processed by EHRbase?
[...]

PR #522, opened 2021-06-02
Feature/sync/support multi paths in template

chevalleyc:
## Changes

- Support aql query on node with multiple paths within a template. 
- uses `union` instead of `union all` in sql query aggregation to remove duplicate results

NB. Depending on the complexity of templates, 'unqualified' aql query may be slow. F.e. the following query:

```
Select e/ehr_id/value as ehrId, o0/data[at0001]/events[at0002]/data[at0003]/items[at0004]/value/magnitude as F2
 from EHR e 
    contains OBSERVATION o0
      where (e/ehr_id/value = '79c7b669-a067-4f61-91ca-0ef23116d449'
            and 
           o0/data[at0001]/events[at0002]/data[at0003]/items[at0004]/value/magnitude = 1.1)
```
results in a close to 10'000 lines sql query due to the combination of undertemined where clause on `OBSERVATION o` in an environment hosting ~ 10 templates... Although the query items are cached (the first run takes about 10 secs) such a query takes about 1-2 secs which is not reasonable. As a side note, the above filter check conditions on uncomparable data values and is rather meaningless.

It is therefore advised to use properly qualified node with the finer granularity possible in order to reduce the complexity of the resulting SQL query. 

## Related issue

fixes: https://github.com/ehrbase/ehrbase/issues/464
fixes: https://github.com/ehrbase/ehrbase/issues/496
fixes: https://github.com/ehrbase/ehrbase/issues/498

### DIFF
--- a/application/src/main/resources/application-local.yml
+++ b/application/src/main/resources/application-local.yml
@@ -34,9 +34,9 @@ server:
     # if true, WHERE clause is using jsquery, false uses SQL only
     useJsQuery: false
     # ignore unbounded item in path starting with one of
-    ignoreIterativeNodeList: 'activities,content'
+    ignoreIterativeNodeList: 'dummy'
     # how many embedded jsonb_array_elements(..) are acceptable? Recommended == 2
-    iterationScanDepth: 2
+    iterationScanDepth: 20
 
 security:
   authType: NONE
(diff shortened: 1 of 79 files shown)


jakesmolka (approved):
LGTM as discussed. Needs a new separate issue regarding testing strategy of the AQL `dummy` etc. setting though, as discussed too.

PR #535, opened 2021-07-05
create null select for node not defined in template

chevalleyc:
## Changes

This fix force a NULL select (e.g. SELECT 1 ... WHERE 1 = 0) whenever a query refers to a non template defined node. F.e. 

```
SELECT c FROM EHR e CONTAINS COMPOSITION c CONTAINS OBSERVATION
```
applied to a single composition where the template definition doesn't specify such an OBSERVATION. The result then should be NULL and hence empty resultset.

## Related issue

Fixes: https://github.com/ehrbase/project_management/issues/360


ppazos (approved):
This worked as expected. I have tested it with different queries for different entries and the only one retrieving results is the one for ADMIN_ENTRY.

wlad:
comment to myself @wlad 

- [ ] create ticket to add new tests

PR #520, opened 2021-05-31
Fixes ehrscape compo uid

jakesmolka:
## Changes

- EhrScape UID fix

## Related issue

fixes https://github.com/ehrbase/ehrbase/issues/481

## Additional information and checks

- [x] Pull request linked in changelog

PR #552, opened 2021-07-22
Feature/526 audit attributes

jakesmolka:
## Changes

- Modify DB schema to make audit metadata mandatory
- Implement default handling for audit metadata:
    - Default system (was already available)
    - Default committer (read from auth user)
- Refactoring of versioned object (interfaces) on service layer
- Refactoring of versioned object (interfaces) on access layer
- Adaptation of new interfaces in handling of versioned objects:
    - Folder
    - Composition
    - Status (still embedded in EHR access handling)
- BONUS: Fix admin delete of multiple status versions. Found that bug by chance...

## Related issue

https://github.com/ehrbase/project_management/issues/526


rest-ehr-scape/src/main/java/org/ehrbase/rest/ehrscape/controller/CompositionController.java
```diff
@@ -62,11 +65,17 @@ public CompositionController(CompositionService compositionService) {
...
+        Optional<CompositionDto> optionalCompositionDto = compositionService.create(ehrId, compoObj);
+
+        var compositionUuid = optionalCompositionDto.orElseThrow(() ->
+            new InternalServerException("Failed to create composition"))
+            .getUuid();
 
         CompositionWriteRestResponseData responseData = new CompositionWriteRestResponseData();
         responseData.setAction(Action.CREATE);
-        responseData.setCompositionUid(compositionId + "::" + 1);
+        responseData.setCompositionUid(compositionUuid + "::" + 1);
```
ppazos, line 78: @jakesmolka missing the creating_system_id in the OBJECT_VERSION_ID?

https://specifications.openehr.org/releases/UML/latest/index.html#Architecture___18_5_1_83e026d_1503073094521_559965_7682

service/src/main/java/org/ehrbase/dao/access/jooq/ContributionAccess.java
```diff
@@ -180,21 +181,16 @@ public UUID commit(Timestamp transactionTime, UUID committerId, UUID systemId, C
...
-        if (committerId == null) {
-            //get current user from JVM
-            String defaultUser = System.getProperty("user.name");
-            //check for that user in the DB
-            String scheme = System.getProperty("host.name");
-            if (scheme == null)
-                scheme = "local";
-            committerId = new PersistedPartyProxy(this).getOrCreate(defaultUser, UUID.randomUUID().toString(), scheme, getServerConfig().getNodename(), "PARTY");
+        if (committerId != null) {
+            auditDetails.setCommitter(committerId);
```
ppazos, line 185: @jakesmolka shouldn't the committerId variable be a PartyProxy instead of a UUID? Not sure how this is modelled internally. For instance what happens if there is a committer name and identifiers? (if the comitter is PARTY_IDENTIFIED https://specifications.openehr.org/releases/UML/latest/index.html#Diagrams___18_1_83e026d_1433773264580_430294_7408)

service/src/main/java/org/ehrbase/dao/access/jooq/FolderAccess.java
```diff
@@ -110,70 +110,61 @@ public FolderAccess(I_DomainAccess domainAccess, UUID ehrId, I_ContributionAcces
...
+        UUID contributionAccessEhrId = this.contributionAccess.getEhrId();
+        /*save the EHR id from oldContribution since it will be the same as this is an update operation*/
+        if (this.contributionAccess.getEhrId() == null) {
+            ContributionRecord rec = getContext().fetchOne(CONTRIBUTION, CONTRIBUTION.ID.eq(oldContribution));
+            contributionAccessEhrId = rec.getEhrId();
         }
+        this.contributionAccess.setEhrId(contributionAccessEhrId);
+
+        this.contributionAccess.commit(Timestamp.valueOf(transactionTime), committerId, systemId,
+            ContributionDataType.folder, ContributionDef.ContributionState.COMPLETE, I_ConceptAccess.ContributionChangeType.MODIFICATION, description);
```
ppazos, line 134: @jakesmolka is the change type always MODIFICATION here? can't be AMENDMENT?

base/src/main/resources/db/migration/V51__audit_attributes_fix.sql
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+-- Fix mandatory attributes with NOT NULL constraint
+ALTER TABLE ehr.audit_details
+    ALTER COLUMN system_id SET NOT NULL,
+    ALTER COLUMN committer SET NOT NULL,
```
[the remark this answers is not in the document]
jakesmolka: The original column is named `committer`.

base/src/main/resources/db/migration/V52__admin_delete_status_fix.sql
```diff
@@ -0,0 +1,46 @@
...
+-- Description: Function to delete an EHR, incl. Status.
+-- Parameters:
+--    @ehr_id_input - UUID of target EHR
+-- Returns: '1' and linked audit, party UUID
+-- Requires: Afterwards deletion of returned audit and party.
+-- =====================================================================
+CREATE OR REPLACE FUNCTION ehr.admin_delete_ehr(ehr_id_input UUID)
+RETURNS TABLE (num integer, status_audit UUID, status_party UUID) AS $$
+BEGIN
+RETURN QUERY WITH linked_status(has_audit) AS ( -- get linked STATUS parameters
```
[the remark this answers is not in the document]
jakesmolka: The whole Admin API is a bit more complicated. Its implementation is documented here: https://wiki.vitagroup.ag/display/ETHERCIS/Implementation (the page slug is misleading though).
But in short: Yes, all those linked objects are removed too, but not via this function.

rest-ehr-scape/src/main/java/org/ehrbase/rest/ehrscape/controller/CompositionController.java
jakesmolka: Well, yes but no :smile: This is in the scope of the EHRscape API. And on the one hand, this is out of scope as I was just changing the DTOs due to refactoring here, and on the other hand, this particular fix is waiting over there already: https://github.com/ehrbase/ehrbase/pull/520
(FYI: Renaud just filed the fix as PR to Archie yesterday, so #520 should be able to move forward soon.)

But good spotting there, thanks!

service/src/main/java/org/ehrbase/dao/access/jooq/ContributionAccess.java
jakesmolka: Thorough thinking again, thanks! In short: This works with all kinds of PartyProxy (which are implemented).

Longer: Formally, for this refactoring I only removed the creation of default values here, since that is happening centralized now.
But regarding Parties: Here in the access layer this audit's committer is directly corresponding to an entry in the party DB table, hence the UUID. The way the party table and its handling is implemented can handle all the types. Technically it has a `party_type` column, which allows distinct processing, which might include looking for identifiers in the `identifier` table.
For instance, the new default committer is actually a PARTY_IDENTIFIED (see here: https://github.com/ehrbase/ehrbase/pull/552/files/1e1db2070aab5c71f5f7bae1ab2ce995e9eb2c84#diff-005b654bd9e859813f3f0618563b53272587b48c2ed26e8fb2ffba98229a3871R79).

service/src/main/java/org/ehrbase/dao/access/jooq/FolderAccess.java
jakesmolka: Good one, thanks! I'll commit a change.

jakesmolka:
@ppazos with the latest change this PR is ready to go from my side. But keep reviewing if you like.

ppazos:
@jakesmolka I can see the changes from previously reviewed code, commit what you like, no problem.

base/src/main/resources/db/migration/V52__admin_delete_status_fix.sql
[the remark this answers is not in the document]
ppazos: thanks.

service/src/main/java/org/ehrbase/service/FolderServiceImp.java
```diff
@@ -108,20 +117,175 @@ public ObjectVersionId create(UUID ehrId, Folder content, UUID contribution, UUI
...
+        ObjectVersionId folderId;
+        if (contribution == null) {
+            folderId = new ObjectVersionId(
+                folderAccess.commit(LocalDateTime.now(), systemId, committerId, description).toString()
+                    + "::" + getServerConfig().getNodename()
+                    + "::1"
+            );
+        } else {
+            folderId = new ObjectVersionId((folderAccess.commit(LocalDateTime.now(), contribution).toString()
+                + "::" + getServerConfig().getNodename()
```
ppazos, line 141: @jakesmolka Are we are overwriting the values sent by the client for the object version id? For instance, for the creating_system_id what happens if the client sends something different than the value we have configured? https://specifications.openehr.org/releases/ITS-REST/latest/ehr.html#contribution-contribution-post

service/src/main/java/org/ehrbase/service/BaseServiceImp.java
```diff
@@ -0,0 +1,97 @@
...
+    var identifier = new DvIdentifier();
+    identifier.setId(name);
+    identifier.setIssuer("EHRbase");
+    identifier.setAssigner("EHRbase");
+    identifier.setType("EHRbase Security Authentication User");
+    identifiers.add(identifier);
+    // Following getOrCreate will check for matching party with given UUID, but as it is random, it also checks
+    // for matching name + identifiers. So it will find already created parties for existing users.
+    return new PersistedPartyProxy(getDataAccess())
+        .getOrCreate("EHRbase Internal " + name, UUID.randomUUID().toString(), DEMOGRAPHIC, "User",
```
ppazos, line 89: @jakesmolka what's the role of the namespace "User"? Is this related to a security feature?

service/src/main/java/org/ehrbase/service/FolderServiceImp.java
jakesmolka: Interesting question again! But this is kind of out of scope of this PR. Anyway, here are my thoughts on that:

In EHRbase, if a client commits an object via direct or contribution endpoint, the given UID will be ignored and handled server-side.

I'm talking about what's illustrated in that little composition snippet:

``` json
{
  "_type": "COMPOSITION",
  ...
  "uid": {
    "_type": "OBJECT_VERSION_ID",
    "value": "8849182c-82ad-4088-a07f-48ead4180515::openEHRSys.example.com::1"
  },
  ...
```

Overall there's some discussion in the SEC, because this behavior is not really 100% clear in the API docs. See: https://openehr.atlassian.net/browse/SPECITS-62 
Please comment your perspective over there :slightly_smiling_face: 

And specifically regarding the "creating_system_id": EHRbase doesn't support imported versions or alike yet. So there's no concept of data from other systems. Data committed (especially like in this case via initial object creation) will be treated to be originally created in the local EHRbase instance. Therefore it always used the current internal system ID.

service/src/main/java/org/ehrbase/service/BaseServiceImp.java
jakesmolka: Good question! The original implementation used it already. See:
https://github.com/ehrbase/ehrbase/blob/develop/service/src/main/java/org/ehrbase/service/BaseService.java#L58

``` java
protected UUID getUserUuid() {
        //@TODO READ from Spring Security
        return new PersistedPartyProxy(getDataAccess()).getOrCreate(null, "cbf741ff-9480-4792-8894-13fc5f818b6d", DEMOGRAPHIC, "User", PARTY);
    }
```

I'm not sure what makes sense here. Do you have a suggestion?

service/src/main/java/org/ehrbase/dao/access/jooq/FolderAccess.java
```diff
@@ -110,70 +110,61 @@ public FolderAccess(I_DomainAccess domainAccess, UUID ehrId, I_ContributionAcces
...
+        UUID contributionAccessEhrId = this.contributionAccess.getEhrId();
+        /*save the EHR id from oldContribution since it will be the same as this is an update operation*/
+        if (this.contributionAccess.getEhrId() == null) {
+            ContributionRecord rec = getContext().fetchOne(CONTRIBUTION, CONTRIBUTION.ID.eq(oldContribution));
+            contributionAccessEhrId = rec.getEhrId();
         }
+        this.contributionAccess.setEhrId(contributionAccessEhrId);
+
+        this.contributionAccess.commit(Timestamp.valueOf(transactionTime), committerId, systemId,
+            ContributionDataType.folder, ContributionDef.ContributionState.COMPLETE, changeType, description);
```
ppazos, line 134: @jakesmolka I've checke the ContributionsAccess and I'm not sure I understand the role of the ContributionState, since the VERSION is the one that has a lifecycle_state, not the CONTRIBUTION, maybe it's an internal thing for the persistence to have this COMPLETE state? REF https://specifications.openehr.org/releases/RM/Release-1.1.0/common.html#_overview_5

Also if you check the code of the commit methods, internally it is using COMPLETE if the parameter is null.

service/src/main/java/org/ehrbase/dao/access/jooq/FolderAccess.java
```diff
@@ -511,17 +495,35 @@ public static I_FolderAccess getNewFolderAccessInstance(final I_DomainAccess dom
...
-    public Integer delete() {
-        // not valid, because missing mandatory data, like committer
-        throw new InternalServerException(CALLED_INVALID_ACCESS_LAYER_METHOD);
+    public int delete(LocalDateTime timestamp, UUID systemId, UUID committerId, String description) {
+        // create new contribution for this deletion action (with embedded contribution.audit handling)
+        contributionAccess = I_ContributionAccess.getInstance(getDataAccess(),
+            contributionAccess.getEhrId()); // overwrite old contribution with new one
+        var contribution = contributionAccess
+            .commit(TransactionTime.millis(), committerId, systemId, null,
+                ContributionDef.ContributionState.COMPLETE,
```
ppazos, line 509: @jakesmolka not sure if the ContributionState represents the VERSION.lifecycle_state of the VERSION inside the CONSITRIBUTION. If so, and this is a delete operation, the lifecycle_state should be DELETED. REF https://github.com/openEHR/terminology/blob/master/openEHR_RM/en/openehr_terminology.xml#L174

service/src/main/java/org/ehrbase/dao/access/jooq/FolderAccess.java
jakesmolka: Your two newer comments are very good! I think we need to create one or two issues out of them. But again, for this PR this is unfortunately out of scope. As I stated in the beginning, most of the "added" code here is technically just moved, due to the new interfaces :slightly_smiling_face: 

Regarding your second point: True, but we agreed on not supporting states (i.e. in the direction of drafts) just yet. So the method kind of overwrites the null value. But I'm not sure what the authors intention was, why it was necessary to handle a null value and so on.

Your first point is a really good observation. You just discovered a shortcoming in the current implementation we need to address somehow. Currently it is certainly possible that the persisted data represents the state wrong, especially in cases where one contribution has many versioned objects as payload. Then each version would only have one state, the one of the linked contribution.
I'll create an issue for that.

service/src/main/java/org/ehrbase/dao/access/jooq/FolderAccess.java
jakesmolka: See my previous comment for context.

Apart from that, also a good one. I'll document it in the issue, but won't touch the states right now due to the scope.

service/src/main/java/org/ehrbase/dao/access/jooq/FolderAccess.java
jakesmolka: PS: See https://github.com/ehrbase/project_management/issues/557 and https://github.com/ehrbase/project_management/issues/556 (which was not part of your comment, but it goes in the same direction)

jakesmolka:
As discussed with @ppazos there are no more comments regarding the changes of this PR and it can be merged.

PR #626, opened 2021-10-11
Bugfix/project management/581 delete ehr large db

subigre:
## Changes

Refactor delete ehr using one plpgsql function

## Related issue

Resolves ehrbase/project_management#581


subigre:
@chevalleyc, I think that all indexes are now created but the operation is still a bit long on very large DB. 
I am wondering if we should not implement thies feature using batch and async processing?

PR #632, opened 2021-10-14
Feature/remove unused test data sets

wlad:
## Changes

- removed all test-dat-sets which are not use in any of the robot tests


ppazos:
@wlad the time window tests could be removed since that keyword was removed from the specs.

Issue #643, opened 2021-10-20
Commit compo validation error when terminology is local and there is no list provided


ppazos:
## Configuration information

```
<status>
  <ehrbase_version>0.17.2</ehrbase_version>
  <openehr_sdk_version>30e3c81</openehr_sdk_version>
  <archie_version>1.0.3</archie_version>
  <jvm_version>Oracle Corporation 11.0.6+10</jvm_version>
  <os_version>Linux amd64 4.15.0-91-generic</os_version>
  <postgres_version>PostgreSQL 10.14 (Ubuntu 10.14-0ubuntu0.18.04.1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0, 64-bit</postgres_version>
</status>
```

## Steps to reproduce ## 

The OPT test all types for the node at0005 has a DV_CODED_TEXT constraint with terminology_id=local and no list of codes provided.

https://github.com/ehrbase/ehrbase/blob/develop/tests/robot/_resources/test_data_sets/valid_templates/all_types/Test_all_types.opt#L361-L409

With Thomas we think the OPT with that kind of constraint should be treated as invalid, since 'local' means the code SHOULD be in the template. See the discussion:

https://discourse.openehr.org/t/is-terminology-id-local-compatible-to-not-specifying-a-code-list-in-the-archetype/1921

This can be checked via Robot: (remove -v nodocker if you are using docker)

`$ robot -i AQL_loaded_db -e TODO -e FUTURE -d results -L TRACE -v nodocker robot`

## Actual result ##

The issue we have is: the template validates when uploaded, and what fails is the commit of the COMPOSITION, since it has a random code in the DV_CODED_TEXT.

```
{
[...]

PR #644, opened 2021-10-21
Feature/527 test data sets review

ppazos:
## Changes

- missing archetype_details for archetype root nodes
- fixes for intervals of count, quantity and datetime (still need to check interval of time and interval of date, and other possible DV_INTERVAL<T> types)
- fix for xmls which had constraints for the LOCATABLE.name, the constraint wasn't considered
- fix for elements with no constraints, to add dummy null_flavour to validate invariant (ELEMENT.value XOR ELEMENT.null_flavour)
- issue with RIPPLE conformance template, the proportion has is_integral as an attribute but it is a function in the RM, asked the SEC, Thomas agreed in manually remove the constraint from the OPT (this particular template was the one making the compo invalid in the Postman collection, plus other fixes above also were happening in the POST /compositon in the Postman collection)
- When CLUSTER has no constraints in the children or only SLOTs in the children, no items were generated, and CLUSTER.items is 1..* in the RM causing a validation issue, now a dummy ELEMENT is added in any of these cases.
- detected validity issue in test_all_types template, fixed the issue in the template, regenerated the composition data set and raised https://github.com/ehrbase/ehrbase/issues/643


wlad (approved):
@ppazos please see my https://github.com/ehrbase/ehrbase/issues/643#issuecomment-975483623. Apart from that everything looks good :+1:

Issue #628, opened 2021-10-12
little regression in AQL response


wlad:
## Test Case/s To Reproduce Issue

D-500 Execute Ad-HOc Query - Get Data

```
# path to test case
tests/robot/QUERY_SERVICE_TESTS/A.1_EXECUTE_AD-HOC_QUERY/A.1__a_Loaded_DB.robot
```

```
# robot command to execute related test case(s) in your terminal/console

robot -i Aql_loaded_dbAND628 -d results -L TRACE robot/QUERY_SERVICE_TESTS


```

## Actual Result (left) | Expected Result (right)

![image](https://user-images.githubusercontent.com/64201310/136918414-2dbdc46c-aaa1-4647-92c5-8a96df2a558c.png)

PR #645, opened 2021-10-21
Feature/628 little regression

chevalleyc:
## Changes

Ensure that whenever no alias is provided, it leads to this field numbering we have with the REST api (that is: #0, #1, etc.)

## Related issue

fixes: https://github.com/ehrbase/ehrbase/issues/628


wlad (approved):
Retest passes after fixing some issue with new test-data-sets.
Sonar complains about 60.8% Duplicated Lines (%) on New Code (is greater than 20%), but I guess we can ignore that for now.

PR #680, opened 2021-11-25
Pem 316 enhanced caching

subigre:
## Changes

Migrate EHRbase code from JCache API to Spring Cache.

## Related issue

PEM-316

### DIFF
--- a/service/src/main/java/org/ehrbase/service/TemplateDBStorageService.java
+++ b/service/src/main/java/org/ehrbase/service/TemplateDBStorageService.java
@@ -20,89 +20,83 @@
 
 package org.ehrbase.service;
 
+import java.util.List;
+import java.util.Optional;
 import java.util.Set;
-import java.util.stream.Collectors;
 import org.ehrbase.api.definitions.ServerConfig;
 import org.ehrbase.dao.access.interfaces.I_DomainAccess;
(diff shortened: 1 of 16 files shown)


service/src/main/java/org/ehrbase/service/TemplateDBStorageService.java
```diff
@@ -48,14 +51,20 @@ public TemplateDBStorageService(DSLContext context, ServerConfig serverConfig) {
...
     }
 
+    // Extracts and stores the template ID in a separate column?
+    @Override
+    public Set<String> findAllTemplateIds() {
+        return listAllOperationalTemplates()
+            .stream()
+            .map(metadata -> TemplateUtils.getTemplateId(metadata.getOperationaltemplate()))
+            .collect(Collectors.toSet());
+    }
```
stefanspiska, line 64: // Extracts and stores the template ID in a separate column?

Yes please otherwise this will be super slow.

stefanspiska: There is already a Collum template_id in the db

stefanspiska (changes requested):

stefanspiska (approved):

PR #712, opened 2021-12-16
Replaced TOP with LIMIT in aql queries

adilmushtaq:
CDR-32


tests/robot/_resources/test_data_sets/query/expected_results/loaded_db/C/101_get_entries_top_5.json
```diff
@@ -1,5 +1,5 @@
 {
-  "q": "SELECT TOP 5 entry FROM EHR e [ehr_id/value='__MODIFY_EHR_ID_1__'] CONTAINS COMPOSITION c CONTAINS ENTRY entry",
+  "q": "SELECT entry FROM EHR e [ehr_id/value='__MODIFY_EHR_ID_1__'] CONTAINS COMPOSITION c CONTAINS ENTRY entry LIMIT 5 ORDER BY entry",
```
stefanspiska, line 2: entry is a json file so order by dos not make much sense here. 

we could remove ORDER BY, but then it would be just random 5 elements. 

I would check way we have even the limit /top in in the fist place.

same for the other test with ORDER BY entry

stefanspiska (changes requested):

tests/robot/_resources/test_data_sets/query/expected_results/loaded_db/C/101_get_entries_top_5.json
adilmushtaq: changes done

stefanspiska (approved):

PR #723, opened 2022-01-11
Feature/cdr 18

chevalleyc:
## Changes
Make sure that table `entry` is used in the FROM clause whenever the template is not resolved (that is resulting in "null" field).

## Related issue

fixes: https://jira.vitagroup.ag/browse/CDR-18

### DIFF
--- a/service/src/main/java/org/ehrbase/aql/sql/QueryProcessor.java
+++ b/service/src/main/java/org/ehrbase/aql/sql/QueryProcessor.java
@@ -35,6 +35,7 @@ import org.ehrbase.aql.sql.queryimpl.attribute.JoinSetup;
 import org.ehrbase.dao.access.interfaces.I_DomainAccess;
 import org.ehrbase.service.IntrospectService;
 import org.jooq.*;
+import org.jooq.Record;
 import org.jooq.conf.Settings;
 import org.jooq.impl.DSL;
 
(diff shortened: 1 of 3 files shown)


service/src/main/java/org/ehrbase/aql/sql/QueryProcessor.java
```diff
@@ -35,6 +35,7 @@
 import org.ehrbase.dao.access.interfaces.I_DomainAccess;
 import org.ehrbase.service.IntrospectService;
 import org.jooq.*;
+import org.jooq.Record;
```
subigre, line 38: Minor but we should either go with wildcard imports or either remove them but I think that we have to keep consistency in the code.

From personal point of view, I prefer the second option. In addition, it is compliant with Google Java style guide.

subigre (commented):
The PR is not mentioned in `CHANGELOG.md `

service/src/main/java/org/ehrbase/aql/sql/QueryProcessor.java
chevalleyc: I did this change as it was causing a compile error with the IDE (but not with Maven...). Basically, it was conflicting with a java.lang.Record. Not quite sure what is the issue TBH

subigre: You are right (One more reasons to avoid wildcard imports 😛 )

subigre (approved):

PR #772, opened 2022-03-18
Feature/add plugin system

stefanspiska:


plugin/src/main/java/org/ehrbase/plugin/EhrBasePlugin.java
```diff
@@ -0,0 +1,41 @@
...
+ * @author Stefan Spiska
+ */
+public abstract  class EhrBasePlugin extends SpringPlugin {
+
+  protected EhrBasePlugin(PluginWrapper wrapper) {
+        super(wrapper);
+    }
+
+    @Override
+    protected ApplicationContext createApplicationContext() {
```
vmueller-vg, line 34: The formatting seems to be a little off here

vmueller-vg, line 34: The formatting seems to be a little off here

service/src/main/java/org/ehrbase/plugin/PluginAspect.java
```diff
@@ -0,0 +1,180 @@
...
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * @author Stefan Spiska
+ */
+@Component
+@Aspect
+public class PluginAspect {
```
vmueller-vg, line 40: @ConditionalOnProperty(prefix = PLUGIN_MANAGER_PREFIX, name = "enable", havingValue = "true")
should be added here, since we do not need this aspect in the invocation chain if we do not use plugins

vmueller-vg, line 40: @ConditionalOnProperty(prefix = PLUGIN_MANAGER_PREFIX, name = "enable", havingValue = "true")
should be added here, since we do not need this aspect in the invocation chain if we do not use plugins

service/src/main/java/org/ehrbase/plugin/PluginAspect.java
```diff
@@ -0,0 +1,180 @@
...
+    return Optional.of(
+        handleChain(
+            chain,
+            l -> (l::aroundCreation),
+            input,
+            i -> {
+              pjp.getArgs()[1] = i.getComposition();
+              pjp.getArgs()[0] = i.getEhrId();
+
+              return ((Optional<UUID>) proceed(pjp)).orElseThrow();
```
vmueller-vg, line 92: I think we should switch this to be something like `proceed(pjp, new Object[]{i.getEhrId(),i.getComposition()}))`

This way we would not rely on the Implementation of `ProceedingJoinPoint::getArgs` to always hand out the same instance of he array. (It currently does, but we can not assume it always will)

vmueller-vg, line 92: I think we should switch this to be something like `proceed(pjp, new Object[]{i.getEhrId(),i.getComposition()}))`

This way we would not rely on the Implementation of `ProceedingJoinPoint::getArgs` to always hand out the same instance of he array. (It currently does, but we can not assume it always will)

service/src/main/java/org/ehrbase/plugin/PluginAspect.java
vmueller-vg, line 40: @ConditionalOnProperty(prefix = PLUGIN_MANAGER_PREFIX, name = "enable", havingValue = "true")
should be added here, since we do not need this aspect in the invocation chain if we do not use plugins

service/src/main/java/org/ehrbase/plugin/PluginAspect.java
vmueller-vg, line 92: I think we should switch this to be something like 
`proceed(pjp, new Object[]{i.getEhrId(),i.getComposition()}))`

This way we would not rely on the Implementation of `ProceedingJoinPoint::getArgs` to always hand out the same instance of he array. (It currently does, but we can not assume it always will)

plugin/src/main/java/org/ehrbase/plugin/EhrBasePlugin.java
```diff
@@ -0,0 +1,41 @@
...
+        return getDispatcherServlet().getWebApplicationContext();
+    }
+
+    public abstract DispatcherServlet getDispatcherServlet();
```
vmueller-vg, line 38: Please fix the indentation

vmueller-vg (changes requested):

service/src/main/java/org/ehrbase/plugin/PluginAspect.java
stefanspiska: I do not understand what you mean.

vmueller-vg (approved):

sidharthramesh:
Very cool!

PR #777, opened 2022-03-23
removed auditresultmap from a class member

micha1970:
auditresultmap is now passed as an explicit argument

## Changes

- First change


subigre (commented):
I would suggest to create a Spring bean with the request scope like:
```java
@Component
@RequestScope
public class AuditContext {

     private final Map<String, Object> auditMapResult = new HashMap<>();

     ...
}
```
Currently the map is instantiated in the @RestController, so the behavior is probably good. However, I think there is one drawback with the current approach as it requires to update all signatures and calls where we could simply autowired our "AuditContext" when it is needed.

PR #803, opened 2022-04-06
Feature/cdr 343 Handle Spring MVC Exceptions

vmueller-vg:
## Changes

- Handle ResponseStatusExceptions and 4xx status code related Spring MVC exceptions

## Related issue

- CDR-343


vmueller-vg:
The code duplication value is so high because the two handlers are part of the two different rest interfaces, but they do almost the same... There currently is no good place to put a common superclass.

PR #804, opened 2022-04-07
Handle spring web exceptions

vmueller-vg:
## Changes

- Handle ResponseStatusExceptions and 4xx status code related Spring MVC exceptions

## Related issue

- CDR-343

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


subigre (commented):
Probably not part of this issue but I already raised this problem before and I think that we should think to add `rest-api-common` that might be used as dependency by ``rest-ehr-scape` and `rest-openehr`.
It will facilitate the maintenance and also keep consistency between our APIs (e.g. exception, Location header, etc)

vmueller-vg:
> Probably not part of this issue but I already raised this problem before and I think that we should definitely think to add `rest-api-common` that might be used as dependency by `rest-ehr-scape` and `rest-openehr`. It will facilitate the maintenance and also keep consistency between our APIs (e.g. exception, Location header, etc)

i completely agree, but that should be a separate ticket to extract all common things ;)

subigre (approved):

PR #817, opened 2022-04-13
Sync/cdr 313

micha1970:
## Changes

- First change


application/src/main/resources/application.yml
```diff
@@ -198,7 +192,11 @@ ipf:
 validation:
   external-terminology:
     enabled: false
-
+    provider:
+      fhir:
+        type: FHIR
+        url: https://r4.ontoserver.csiro.au/fhir/
```
subigre, line 198: I would suggest to remove this url by default

subigre (changes requested):

application/src/main/resources/application.yml
micha1970: I would agree, but the url was present in the code before refactoring. Anyway, the validation is disabled yb default

Issue #811, opened 2022-04-10
Consistent compositionUid for APIs


sidharthramesh:
## Configuration information

```
{
	"ehrbase_version": "0.21.0-SNAPSHOT",
	"openehr_sdk_version": "1.19.0-SNAPSHOT",
	"archie_version": "2.0.1",
	"jvm_version": "Oracle Corporation 11.0.14.1+1",
	"os_version": "Linux amd64 5.10.0-0.bpo.9-amd64",
	"postgres_version": "PostgreSQL 14.2 on x86_64-pc-linux-gnu, compiled by gcc, a cf00366d51 p a16b4ae78f, 64-bit"
}
```

## Steps to reproduce ## 

Creating compositions using POST `{{ehrscapeBaseUrl}}/composition?ehrId={{ehrId}}&templateId={{templateId}}&format=FLAT` returns compositionUid in fully versioned format:

```
{
"compositionUid": "fea1c081-252d-4c0f-8e78-8c788395aa8a::ehrbase.course.voiceintern.com::1"
}
```

Reading it using GET `{{ehrscapeBaseUrl}}/composition/{{compositionId}}?format=FLAT` returns only the uuid:
```
{
"compositionUid": "fea1c081-252d-4c0f-8e78-8c788395aa8a"
}
```

## Expected result (Acceptance Criteria) ##
Consistent compositionUid format in both APIs.

## Definition of Done ##

- [ ] The defect is checked by an unit or an integration test (Robot)
- [ ] Merge Request approved
- [ ] Unit tests passed
- [ ] Build without errors
- [ ] Release notes prepared
- [ ] No additional runtime warnings

stefanspiska:
The two apis are part of different  specs and those specs also sets the uuid format. So this is not a bug. To solve this we would need a spec how to commit flat on the openehr api.

PR #825, opened 2022-04-27
Consistent compositionUid for APIs

mhenke-vg:
Fixes https://github.com/ehrbase/ehrbase/issues/811

### DIFF
--- a/rest-ehr-scape/src/main/java/org/ehrbase/rest/ehrscape/controller/CompositionController.java
+++ b/rest-ehr-scape/src/main/java/org/ehrbase/rest/ehrscape/controller/CompositionController.java
@@ -38,176 +38,178 @@ import java.util.UUID;
 
 @RestController
 @RequestMapping(
-    path = "/rest/ecis/v1/composition",
-    produces = {MediaType.APPLICATION_JSON_VALUE, MediaType.APPLICATION_XML_VALUE}
+        path = "/rest/ecis/v1/composition",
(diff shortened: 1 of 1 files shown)


rest-ehr-scape/src/main/java/org/ehrbase/rest/ehrscape/controller/CompositionController.java
```diff
@@ -100,7 +100,7 @@ public ResponseEntity<CompositionResponseData> getComposition(
       responseDto.setAction(Action.RETRIEVE);
       responseDto.setFormat(format);
       responseDto.setTemplateId(compositionDto.get().getTemplateId());
-      responseDto.setCompositionUid(compositionDto.get().getUuid().toString());
+      responseDto.setCompositionUid(compositionUid);
```
stefanspiska, line 103: This only works if compositionUid  from the GET is full, wich is not nessery (https://rest.ehrscape.com/rest/v1/composition/9b9ce329-9475-4be3-987c-eaac3a01d88d?format=Flat) is valid for example
You need to calculate id as compositionDto.get().getUuid()+ "::" + compositionService.getServerConfig().getNodename() + "::"
        + compositionService.getLastVersionNumber(compositionDto.get().getUuid())

stefanspiska (changes requested):

stefanspiska (approved):

PR #826, opened 2022-04-28
Fix duplicate users issue

subigre:
## Changes

- Fix duplicate users issue that makes EHRbase create a new user for each request
- Add Flyway migration to fix the issue 

## Related issue

- CDR-347


subigre:
Note: I am not sure that is a good idea to include the PL/pgSQL procedure as Flyway migration as the execution could really take some time for system that runs since a while.

stefanspiska (approved):

PR #864, opened 2022-05-18
Add spotless plugin, Add codestyle check to workflows

vmueller-vg:
## Changes

- Add spotless maven plugin
- add spotless:check as check for PRs

## Related issue
- CDR-385

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -202,6 +202,18 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>com.diffplug.spotless</groupId>
+                <artifactId>spotless-maven-plugin</artifactId>
+                <version>2.22.5</version>
+                <configuration>
+                    <java>
+                        <palantirJavaFormat>
+                            <version>2.24.0</version>
+                        </palantirJavaFormat>
+                    </java>
+                </configuration>
+            </plugin>
         </plugins>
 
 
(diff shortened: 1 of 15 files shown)


pom.xml
```diff
@@ -202,6 +203,25 @@
...
+                            <off>@format:off</off>
+                            <on>@format:on</on>
+                        </toggleOffOn>
+                        <palantirJavaFormat>
+                            <version>2.24.0</version>
+                        </palantirJavaFormat>
+                        <licenseHeader>
+                            <file>${license-header.dir}/license-header</file>
+                        </licenseHeader>
+                    </java>
```
stefanspiska, line 222: you should add a https://github.com/diffplug/spotless/tree/main/plugin-maven#how-can-i-enforce-formatting-gradually-aka-ratchet so since otherwise now everting fails since the old ones are not reformatted.

stefanspiska (changes requested):

pom.xml
vmueller-vg: This is intentional. a full reformat of the project will be committed to develop directly. so for all branches from that point on this will not be a problem. all currently existing branches can rebase to the state after this PR and then do a full reformat themself or merge in develop

stefanspiska: did you check that the format passes?

stefanspiska (approved):

PR #869, opened 2022-05-20
fix release action

stefanspiska:


.github/workflows/release.yml
```diff
@@ -104,7 +134,7 @@ jobs:
           #update sdk to latest snapshot
           mvn versions:update-properties  -Dincludes=org.ehrbase.openehr.sdk:*  -DallowSnapshots=true
           #edit changelog
-          sed -i '5i ## [unreleased]\n ### Added \n ### Fixed \n' CHANGELOG.md
+          sed -i '7i ## [unreleased]\n ### Added\n ### Changed \n ### Fixed \n' CHANGELOG.md
```
HolgerReiseVSys, line 137: Should it be "8i" ?

.github/workflows/release.yml
```diff
@@ -44,7 +69,12 @@ jobs:
           mvn versions:set -DnewVersion=${VERSION} -DprocessAllModules=true
           #Update sdk version latest release 
           mvn versions:update-properties  -Dincludes=org.ehrbase.openehr.sdk:*
+          sdk_version=$( grep -oPm1 "(?<=<ehrbase.sdk.version>)[^<]+" "bom/pom.xml")
+          echo ${sdk_version}
           #edit changelog
+          replace="0,/Changed/{s/Changed/Changed \n - Upgrade openEHR_SDK to version ${sdk_version} see https:\/\/github.com\/ehrbase\/openEHR_SDK\/blob\/develop\/CHANGELOG.md/}"
+          echo "${replace}"
```
HolgerReiseVSys, line 76: is the echo needed?

HolgerReiseVSys (commented):
Please don't take this for a full review

.github/workflows/release.yml
stefanspiska: no. I tested it 
https://github.com/stefanspiska/ehrbase/blob/develop/CHANGELOG.md?plain=1

HolgerReiseVSys: To me it looks like line 7 should be a blank line instead of line 17

vmueller-vg: The double blank line is not produced by this script.

vmueller-vg: Also @HolgerReiseVSys you are rigth it should be 8i if we assume line 7 to be blank. which actually produced the one time double blank line

vmueller-vg (approved):

vmueller-vg (approved):

PR #911, opened 2022-06-14
Export of the connection variables to the DB that are in the pom to a properties file

agustinpallares94:
Export of the connection variables to the DB that are in the pom to a properties file to only have to modify them there and in case of deployment in kubernetes to be able to use secrets

## Changes

1- Create a abc.properties in the main folder with the bbdd variables
2- In the main pom I change the DB connection variables with the variables I have in my abc.properties file.
3- I add the plugin org.codehaus.mojo indicating the directory where the abc.properties file is located.
4- I add this plugin to each child pom only indicating that the path is set by the parent pom.

## Related issue

In this way it is possible to avoid hardcoding the connection to the DB in order to be able to migrate it to kubernetes with secrets in the future. 
Besides, it only allows you to change the .properties data and everything is updated.

This is also very advantageous if you run ehrbase in a pipeline. Normally database data is treated with secrets and only if you have the secrets file you can run the product in the pipeline.

Solution oriented to DevOps philosophy, use of kubernetes and for possible hot deployments. 

### DIFF
--- a/abc.properties
+++ b/abc.properties
@@ -1,0 +1,5 @@
+jdbc.url=localhost
+jdbc.port=5432
+jdbc.db=ehrbase
+jdbc.user=ehrbase
+jdbc.pass=ehrbase
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -76,6 +76,30 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>properties-maven-plugin</artifactId>
+                <version>1.0.0</version>
+                <executions>
+                    <execution>
+                        <phase>initialize</phase>
+                        <goals>
+                            <goal>read-project-properties</goal>
+                        </goals>
+                    </execution>
+                </executions>
+                <configuration>
+                    <files>
+                        <file>${project.parent.basedir}/abc.properties</file>
+                    </files>
+                    <outputFile/>
+                    <properties/>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
 
 
 </project>
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -145,6 +145,28 @@
 
     <build>
         <plugins>
+
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>properties-maven-plugin</artifactId>
+                <version>1.0.0</version>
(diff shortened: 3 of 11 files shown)


stefanspiska:
@agustinpallares94 

the values in the pom are for the build of ehrbase. 
the db connection during runtime is controlled by  https://github.com/ehrbase/ehrbase/blob/develop/application/src/main/resources/application.yml
so since this uses standard spring Mechanismen this can be overwritten by in kubernetes  via

``` yml
  - name: SPRING_DATASOURCE_URL
     value: jdbc:postgresql://10.216.24.64:5432/ehrbase
  - name: DB_USER
     value: ehrbase
  - name: DB_PASS
     valueFrom:
          secretKeyRef:
                name: ha-postgres
                key: password
```


And if you wand to overwrite the properties for a build you can simple use 
mvn clean install -Ddatabase.host=localhost

agustinpallares94:
> @agustinpallares94
> 
> the values in the pom are for the build of ehrbase. the db connection during runtime is controlled by https://github.com/ehrbase/ehrbase/blob/develop/application/src/main/resources/application.yml so since this uses standard spring Mechanismen this can be overwritten by in kubernetes via
> 
> ```yaml
>   - name: SPRING_DATASOURCE_URL
>      value: jdbc:postgresql://10.216.24.64:5432/ehrbase
>   - name: DB_USER
>      value: ehrbase
>   - name: DB_PASS
>      valueFrom:
>           secretKeyRef:
>                 name: ha-postgres
>                 key: password
> ```
> 
> And if you wand to overwrite the properties for a build you can simple use mvn clean install -Ddatabase.host=localhost

Oh Nice!
In the case that you want to build ehrbase in a pipeline using secrets for the db pass, how do you do with this conf?

stefanspiska:
@agustinpallares94 
well in our pipeline we have the CI runner  mount the secret as environments an then there can be simple used in the commands.

see for example for github

```yaml
      - name: run
        run: mvn clean install -Ddatabase.host=${HOST}
        env:
          HOST: ${{ secrets.DB_HOST }}

```

agustinpallares94:
> @agustinpallares94 well in our pipeline we have the CI runner mount the secret as environments and then there can be simple used in the commands.
> 
> see for example for github
> 
> ```yaml
>       - name: run
>         run: mvn clean install -Ddatabase.host=${HOST}
>         env:
>           HOST: ${{ secrets.DB_HOST }}
> ```

Oh okay! than you so much Stefan

PR #971, opened 2022-08-25
added redis dependency to POM and optional coniguration parameter to …

markuswagnervitagroup:
…application.yml

## Changes

- First change

### DIFF
--- a/application/src/main/resources/application.yml
+++ b/application/src/main/resources/application.yml
@@ -35,9 +35,10 @@ spring:
 
   cache:
     type: jcache # change to type redis if usage of redis distributed cache is intended
-    jcache:
+    jcache: # jcache parameters are only used if cache.type=jcache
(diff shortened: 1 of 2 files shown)


application/src/main/resources/application.yml
```diff
@@ -34,9 +34,14 @@ spring:
     name: ehrbase
 
   cache:
+    type: jcache # change to type redis if usage of redis distributed cache is intended
     jcache:
       config: classpath:ehcache.xml
 
+  redis:
```
stefanspiska, line 41: please add a comment that those are only used if type redis

stefanspiska (changes requested):

application/src/main/resources/application.yml
markuswagnervitagroup: added comments as requested

PR #984, opened 2022-09-15
Fix/concurent user creation

stefanspiska:


service/src/main/java/org/ehrbase/service/SystemService.java
```diff
@@ -0,0 +1,126 @@
...
+    protected UUID getCurrentUserId() {
+        var username = authenticationFacade.getAuthentication().getName();
+        var existingUser = new PersistedPartyIdentified(getDataAccess()).findInternalUserId(username);
+        if (existingUser.isEmpty()) {
+            return getCurrentUserIdSnyc();
+        }
+        return existingUser.get();
+    }
+
+    private synchronized UUID getCurrentUserIdSnyc() {
```
HolgerReiseVSys, line 95: Are the transactions somehow contained to the synchronized block?
Overlapping would result in duplicate users.

HolgerReiseVSys (changes requested):

vidi42:
Internal tracking: https://jira.vitagroup.ag/browse/CDR-547

PR #993, opened 2022-11-08
Consider build without having PostgreSQL installed and running on developers machine or CI-system

jbellmann:
## Changes

use [maven-docker-plugin](https://github.com/fabric8io/docker-maven-plugin)
- to start needed postgresql or code-generation in module 'jooq-pg'
- ~~to start needed postgresql during integration-test phase in module 'application'~~ (Not needed anymore, all tests in that module have been marked with `@Disabled` by someone )

## Related issue

- no related issues (afaik)

In my opinion this could improve 'first-contact-with-ehrbase-experience'.
Just clone, easy build without any setup needed (please also have a look to this PR for using [maven-wrapper](https://github.com/ehrbase/ehrbase/pull/992)   ;-) ), and with a proper successful build importing the project into any IDE works like a charm.


flowrider3000:
How about using testcontainers? 🤔

jbellmann:
Hi Flo,
thanks for asking that question.

With regards to this PR I would say testcontainers is not needed/useful in this context.

The module that is touched here (`jooq-pq`) is only about generating code by JOOQs code generator. There isn't one test written/coded in this module. That's why imo the best option for this module is to go with just taking what is already available and easy to setup/configure in the pom file. Assemble some plugins that will do the job. Done.

With the whole project in mind and writing useful tests by developers you are absolutely right. Testcontainers will be a very good option. I experimented a bit with that in my `ehrbase` fork.

I created a [PostgresInitializer in a single shareable module](https://github.com/jbellmann/ehrbase/blob/modularize/modules/openehr-test-support-postgres/src/main/java/org/ehrbase/test/database/PostgresInitializer.java).  And reuse it in [multiple tests](https://github.com/jbellmann/ehrbase/blob/modularize/modules/openehr-query/src/test/java/org/ehrbase/modules/query/QueryRepositoryTest.java#L16) where a [database is needed](https://github.com/jbellmann/ehrbase/blob/modularize/modules/openehr-query-autoconfiguration/src/test/java/org/ehrbase/modules/query/QueryServiceTest.java#L30) by just specify [usage of that initializer](https://github.com/jbellmann/ehrbase/blob/modularize/modules/openehr-tenant/src/test/java/org/ehrbase/modules/tenant/TenantRepositoryTest.java#L31).

With dependencies available with scope `test` like `flyway-core` and `base` (transient dependency via `jooq-pq`) the database is set up with the latest migrations available in the project.

That could help getting [these kind of test working](https://github.com/ehrbase/ehrbase/blob/develop/application/src/test/java/org/ehrbase/application/abac/AbacIntegrationTest.java#L64). But the real issues there are not only the database. There is more to do. I tried it ;-)

flowrider3000:
Hey Jörg,
Thx for the explanation. I have not recognized, that the PR touches only the `jooq-pq` module.

PR #1020, opened 2022-12-14
fix: removing non required trigger on audit details table

robert-hunter-nes:
## Changes

Updating the `migrate_to_cloud_db_setup.sql` script to so that it doesn't create a `versioning_trigger` on the `ehr.audit_details` table. There is not a history table (`ehr.audit_details_history`) so it appears like the trigger here is incorrect. 

## Related issue

There's no related issue. We had run the script which resulted in us having a trigger on the `audit_details` table that was causing us an error when we tried to create an EHR as the trigger couldn't write to the non-existent `ehr.audit_details_history` table.

I'm not sure we need to include anything in the changelog here. The change is just to avoid anyone else running the migration script and being affected by the same issue.

### DIFF
--- a/base/db-setup/migrate_to_cloud_db_setup.sql
+++ b/base/db-setup/migrate_to_cloud_db_setup.sql
@@ -24,12 +24,6 @@ EXECUTE PROCEDURE ext.versioning('sys_period', 'ehr.folder_items_history', 'true
 
 CREATE TRIGGER versioning_trigger
   BEFORE INSERT OR UPDATE OR DELETE
-  ON ehr.audit_details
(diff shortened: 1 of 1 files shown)


robert-hunter-nes:
I should have asked, is it OK to create a PR without raising an issue? If you would like me to raise one, please just let me know.

stefanspiska (approved):

stefanspiska:
> I should have asked, is it OK to create a PR without raising an issue? If you would like me to raise one, please just let me know.

@robert-hunter-nes this would normally be better.

PR #1024, opened 2022-12-16
validate if party already exist for ehr

dmitrytulba:
## Changes

- First change


service/src/main/java/org/ehrbase/service/EhrServiceImp.java
```diff
@@ -84,6 +87,8 @@
 @Transactional()
 public class EhrServiceImp extends BaseServiceImp implements EhrService {
     public static final String DESCRIPTION = "description";
+    public static final String EHR_NOT_FOUND_WITH_SUPPLIED_SUBJECT =
```
vidi42, line 90: We should maybe change the message to say that the `partyId` is used by a different EHR in the same `partyNamespace`

vidi42 (changes requested):

vidi42 (approved):

PR #1026, opened 2022-12-23
Feature/cdr 648

micha1970:
## Changes

- First change

### DIFF
--- a/plugin/src/main/java/org/ehrbase/plugin/security/PluginSecurityConfiguration.java
+++ b/plugin/src/main/java/org/ehrbase/plugin/security/PluginSecurityConfiguration.java
@@ -1,0 +1,157 @@
+/*
+ * Copyright (c) 2022 vitasystems GmbH and Hannover Medical School.
+ *
+ * This file is part of project EHRbase
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.ehrbase.plugin.security;
+
+import java.lang.reflect.Method;
+import java.util.List;
+import java.util.Objects;
+
+import org.aopalliance.intercept.MethodInterceptor;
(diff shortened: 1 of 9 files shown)


plugin/src/main/java/org/ehrbase/plugin/security/PluginSecurityConfiguration.java
```diff
@@ -0,0 +1,156 @@
...
+            TenantAware annotation =  Objects.requireNonNull(declaringClass.getAnnotation(TenantAware.class));
+            return getAspect().action(new ProceedingJoinPointAdapter(invocation), List.of(annotation));
+          }
+        });
+  }
+
+  public void setApplicationContext(ApplicationContext applicationContext) throws BeansException {
+    this.applicationContext = applicationContext;
+  }
+  // @format:on
```
vmueller-vg, line 155: please move this to Line 90. For the adapter definitions its fine, but it should not exclude the entire class from formatting. Also please make sure that at least the indentation matches.

service/src/main/java/org/ehrbase/tenant/TenantAspect.java
```diff
@@ -28,15 +29,14 @@
...
-import org.ehrbase.api.tenant.ExtractionStrategyAware;
 import org.ehrbase.api.tenant.TenantAuthentication;
 import org.ehrbase.api.tenant.TenantIdExtractionStrategy;
 import org.springframework.aop.aspectj.MethodInvocationProceedingJoinPoint;
 import org.springframework.security.core.context.SecurityContext;
 import org.springframework.security.core.context.SecurityContextHolder;
 
 @Aspect
-public class TenantAspect implements ExtractionStrategyAware {
+public class TenantAspect implements org.ehrbase.api.aspect.TenantAspect {
```
vmueller-vg, line 39: Interface and implementation class should not have the same name. Since this is the default impl for the TenantAspect i suggest renaming this class to DefaultTenantAspect or something like that.

vmueller-vg (changes requested):

vmueller-vg (changes requested):
Also: Merge or rebase develop and resolve the conflict

plugin/src/main/java/org/ehrbase/plugin/security/PluginSecurityConfiguration.java
micha1970: Done

service/src/main/java/org/ehrbase/tenant/TenantAspect.java
micha1970: Renamed it to Default......

micha1970:
> Also:
> 
> * Merge develop or rebase  to develop and resolve the conflict.
> * update CHANGELOG.md

micha1970:
Mistake

PR #1089, opened 2023-03-29
Respect the predefined value of the composition UID when creating new compositions

stefanspiska:


sidharthramesh:
great feature 👍

This however will still not allow creating a new composition with a UUID using the PUT method right?

service/src/main/java/org/ehrbase/service/CompositionServiceImp.java
```diff
@@ -164,6 +165,37 @@ private UUID createInternal(
             throw new InternalServerException(e);
         }
 
+        if (composition.getUid() instanceof ObjectVersionId objectVersionId) {
+
+            if (objectVersionId.isBranch()) {
+                throw new PreconditionFailedException(
+                        "Version branching is not supported: %s".formatted(objectVersionId.getValue()));
+            }
+            if (Integer.parseInt(objectVersionId.getVersionTreeId().getValue()) != 1) {
```
HolgerReiseVSys, line 174: Could be simpler: First check
`"1".equals(objectVersionId.getVersionTreeId().getValue()`
and then 
`objectVersionId.isBranch()` 
to see what is wrong

HolgerReiseVSys (changes requested):
Looks good, just few niceties

HolgerReiseVSys (approved):

PR #1158, opened 2023-08-02
Feature/cdr 920 templates with same

dmitrytulba:
# Changes

> Describe your changes in a short and concise list.

# Related issue

> Reference related issues, and use one of the [closing keywords, e.g. closes or fixes](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) to link the corresponding issue, if any.

# Additional information 

> Provide additional information for this change, if needed.

# Pre-Merge checklist

- [ ] New code is tested
- [ ] Changelog is updated
- [ ] Code has been reviewed


service/src/main/java/org/ehrbase/service/TemplateDBStorageService.java
```diff
@@ -51,23 +53,33 @@ public Set<String> findAllTemplateIds() {
...
+    }
+
     @Override
     public void storeTemplate(OPERATIONALTEMPLATE template, Short sysTenant) {
-        if (readOperationaltemplate(template.getTemplateId().getValue()).isPresent()) {
-            I_TemplateStoreAccess.getInstance(getDataAccess(), template, sysTenant)
-                    .update();
+        if (readOperationalTemplate(template.getTemplateId().getValue()).isPresent()) {
+            findUuidByTemplateId(TemplateUtils.getTemplateId(template))
+                    .ifPresent(uuid -> I_TemplateStoreAccess.getInstance(getDataAccess(), template, sysTenant, uuid)
```
HolgerReiseVSys, line 65: readOperationalTemplate is not necessary, use findUuidByTemplateId to begin with (+ ifPresentOrElse).

service/src/main/java/org/ehrbase/dao/access/jooq/TemplateStoreAccess.java
```diff
@@ -44,12 +44,20 @@
...
 import org.openehr.schemas.v1.OPERATIONALTEMPLATE;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class TemplateStoreAccess extends DataAccess implements I_TemplateStoreAccess {
+    public static final String COULD_NOT_RETRIEVE_TEMPLATE_FOR_TEMPLATE_ID =
+            "Could not retrieve template for template id:";
+    public static final String EXCEPTION = " exception:";
+
+    private static final Logger logger = LoggerFactory.getLogger(TemplateStoreAccess.class);
```
HolgerReiseVSys, line 60: Is this used?

HolgerReiseVSys (changes requested):

HolgerReiseVSys (changes requested):
I'm not sure if it's worth the effort, but I would prefer to keep the caches consistent.
This means, that if a template is added to a cache, it is added to all caches. If it is removed, it is also removed from all. This means that cache updates would need to be centralized. We could not rely on ::computeIfAbsent, but e.g. if the UUID is not in the cache, we load the template from the DB and call a method that takes care of adding it to all caches.

HolgerReiseVSys (approved):

PR #1308, opened 2024-05-08
Extract junit collection from build and test workflow

alexlehn:
# Changes

Extract Junit test collection into a dedicated workflow that is triggered using `workflow_run` so that PRs from forked repositories shall be able to collect the test results. 

Furthermore all workflow names are adjusted. As a side effect the status wokflow should be fixed as well <- @vidi42 this broken workflow was the reason why the release did not work because it is waiting for the status.

Sadly we need to merge everything into the `develop` to check if it's working -> https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run

```
Note: This event will only trigger a workflow run if the workflow file is on the default branch.
``` 

# Pre-Merge checklist

- [x] New code is tested
- [ ] Changelog is updated
- [ ] Code has been reviewed


stefanspiska:
### Robot Results
| :white_check_mark: Passed | :x: Failed | :next_track_button: Skipped | Total | Pass % | :stopwatch: Duration |
| --- | --- | --- | :---: | :---: | :---: |
| 1394 | 0 | 139 | 1394 | 100 | 0s |

vidi42 (approved):

PR #1325, opened 2024-05-24
CDR-1391 Reject invalid fetch/offset with HTTP 422

alexlehn:
# Changes

Return HTTP 422 Unproccessable Content in case fetch or offset is defined inside of the aql query and as parameter.

# Pre-Merge checklist

- [x] New code is tested
- [ ] Changelog is updated
- [ ] Code has been reviewed


aql-engine/src/test/java/org/ehrbase/openehr/aqlengine/service/AqlQueryServiceImpTest.java
```diff
@@ -55,4 +101,51 @@ void resolveEhrCompositions(String srcAql, String expectedAql) {
...
     }
+
+    @ParameterizedTest
+    @CsvSource(
+            textBlock =
+                    """
+                5||10||Invalid AQL query: fetch is defined on query 5 and as parameter 10
+                5|20||40|Invalid AQL query: offset is defined on query 20 and as parameter 40
+                5|||30|Invalid AQL query: limit is defined on query 5 and offset as parameter 30
+                |||42|Invalid AQL query: provided offset 42 without a limit
```
alexlehn, line 112: @HolgerReiseVSys there are currently 4 different message used, base on the given combination. Not sure, but possible we could combine them into one generic message.

HolgerReiseVSys: Maybe it would be sufficient to state that because the query contains a LIMIT clause, fetch and offset parameters must not be used

alexlehn: Reduced to 2 messages -> https://github.com/ehrbase/ehrbase/pull/1325/files#diff-92431608f1168dac98145160a7f45a69eeecd53d858d5683646664c86f7a83dfR109-R112

aql-engine/src/main/java/org/ehrbase/openehr/aqlengine/service/AqlQueryServiceImp.java
```diff
@@ -188,29 +188,30 @@ private static AqlQuery buildAqlQuery(AqlQueryRequest aqlQueryRequest) {
...
+                                throw new UnprocessableEntityException(
+                                        "Query contains a LIMIT clause, fetch and offset parameters must not be used");
+                            }
+                        },
+                        () -> Optional.ofNullable(fetchParam).ifPresent(aqlQuery::setLimit));
+
+        // assign offset parameter
+        if (aqlQuery.getOffset() == null) {
+            Optional.ofNullable(offsetParam).ifPresent(aqlQuery::setOffset);
         }
```
HolgerReiseVSys, line 209: This feels overly complex, maybe try the following: 
if limitQuery null: assign from parameters
else: check parameters are null

aql-engine/src/test/java/org/ehrbase/openehr/aqlengine/service/AqlQueryServiceImpTest.java
```diff
@@ -55,4 +101,51 @@ void resolveEhrCompositions(String srcAql, String expectedAql) {
...
+        runQueryTest(aqlLimit, aqlOffset, paramLimit, paramOffset);
+    }
+
+    private void runQueryTest(String aqlLimit, String aqlOffset, String paramLimit, String paramOffset) {
+        // @format:off
+        String query = "SELECT s FROM EHR_STATUS s %s %s".formatted(
+                Optional.ofNullable(aqlLimit).filter(s -> !s.isEmpty()).map(s -> "LIMIT " + s).orElse(""),
+                Optional.ofNullable(aqlOffset).filter(s -> !s.isEmpty()).map(s -> "OFFSET " + s).orElse("")
+        );
+        service().query(new AqlQueryRequest(
```
HolgerReiseVSys, line 143: If ::buildAqlQuery was package protected, the static method could be tested without mocking everything.
If the parameter replacement was extracted into i´ts own method, it could be tested with even fever side effects.

aql-engine/src/main/java/org/ehrbase/openehr/aqlengine/service/AqlQueryServiceImp.java
alexlehn: done

aql-engine/src/test/java/org/ehrbase/openehr/aqlengine/service/AqlQueryServiceImpTest.java
alexlehn: adjusted to test only `buildAqlQuery`

aql-engine/src/main/java/org/ehrbase/openehr/aqlengine/service/AqlQueryServiceImp.java
```diff
@@ -183,34 +183,34 @@ private QueryResultDto queryAql(AqlQueryRequest aqlQueryRequest) {
...
-            if (aqlQuery.getOffset() != null) {
-                throw new InvalidApiParameterException(
-                        "Invalid AQL query: fetch is defined on query %s and as parameter %s"
-                                .formatted(aqlQuery.getOffset(), offset));
-            }
-            aqlQuery.setOffset(offset);
+
+        // assign offset parameter
+        if (aqlQuery.getOffset() == null) {
+            aqlQuery.setOffset(offsetParam);
```
HolgerReiseVSys, line 207: This can be done (unconditionally) in `if (limitQuery == null)`

service/src/main/java/org/ehrbase/service/StatusServiceImp.java
```diff
@@ -33,7 +33,7 @@
 public class StatusServiceImp implements StatusService {
 
     @SuppressWarnings("SpringJavaAutowiredFieldsWarningInspection")
-    @Autowired
+    @Autowired(required = false)
```
HolgerReiseVSys, line 36: The annotation does not have to be removed

HolgerReiseVSys (changes requested):

aql-engine/src/main/java/org/ehrbase/openehr/aqlengine/service/AqlQueryServiceImp.java
alexlehn: ups ... slipped in, was from a local test

service/src/main/java/org/ehrbase/service/StatusServiceImp.java
alexlehn: ups ... slipped in, was from a local test

HolgerReiseVSys (approved):

PR #1333, opened 2024-05-30
CDR-1429 Remove @Header schema that fails in spring doc 2.5.0

alexlehn:
# Changes

Removed `schema = @Schema(implementation = MediaType.class)` from swagger definitions. This option is currently broken in spring doc 2.5.0 and fails with a nullpointer because there is a missing check for `components == null` in the transitive swagger-api.

* bug in 2.2.21 - missing null check -> https://github.com/swagger-api/swagger-core/blob/v2.2.21/modules/swagger-core/src/main/java/io/swagger/v3/core/util/AnnotationsUtils.java#L1387
* fixed in 2.2.22 -> https://github.com/swagger-api/swagger-core/blob/v2.2.22/modules/swagger-core/src/main/java/io/swagger/v3/core/util/AnnotationsUtils.java#L1387-L1389

It would be possible to update the version to 2.2.22, but I checked the output of the the swagger ui with and without the schema - there are identical - so the `schema` can be safely removed. 

# Pre-Merge checklist

- [x] New code is tested
- [ ] Changelog is updated
- [ ] Code has been reviewed


vidi42:
Seems it's missing the CHANGELOG entry

alexlehn:
> Seems it's missing the CHANGELOG entry

fixed -> https://github.com/ehrbase/ehrbase/pull/1333/commits/23c47613094104f5d47839d42224c33d4e7fc9f1

micha1970 (approved):

PR #1342, opened 2024-06-05
CDR-1401 Add Robot tests with Basic and OAUTH2 auth types

vladislavploaia:


tests/keycloak/import/ehrbase-realm-exported.json
```diff
@@ -0,0 +1,2143 @@
...
+      "containerId" : "ca3e0ff6-5282-4d18-b860-8d28d7ac1b86",
+      "attributes" : { }
+    }, {
+      "id" : "e47d350d-a951-4ddb-9923-657b981ce38f",
+      "name" : "tenant-admin",
+      "composite" : false,
+      "clientRole" : false,
+      "containerId" : "ca3e0ff6-5282-4d18-b860-8d28d7ac1b86",
+      "attributes" : { }
+    }, {
```
alexlehn, line 87: we should double check what roles are needed ;).

tests/docker-compose-int-test.yml
```diff
@@ -31,23 +30,28 @@ services:
...
+    networks:
+      - ehrbase-net
 
   #
   # Integration tests will be ready after ehrbase is healthy.
   # Will run `runRobotTests`
   #
   ehrbase-integration-tests:
-    image: ${EHRBASE_INTEGRATION_TEST_IMAGE:-ehrbase/integration-tests:latest}
+    image: ${EHRBASE_INTEGRATION_TEST_IMAGE:-ehrbase/integration-tests:latest-feature}
```
alexlehn, line 49: should this be changed back to `:latest`

alexlehn (commented):
The Keycloak is now running for all tests, also in case it is not needed. But it seems it does not affect the tests that much.

tests/keycloak/import/ehrbase-realm-exported.json
vladislavploaia: @alexlehn removed all others roles + users and kept the needed only.

tests/docker-compose-int-test.yml
vladislavploaia: @alexlehn will change back to **:latest** once pipeline is passed after cleaning up the roles and users from Keycloak.
Before merging to develop, will change to latest tag.

alexlehn (approved):
@vladislavploaia please squash & merge ;)

PR #1387, opened 2024-08-09
 CDR-1532 Add EHRbase CLI to support pre migrations

alexlehn:
# Changes

Add CLI app mode with commands to run pre migration operations. The CLI mode uses the same spring application stack without WebMVC. This allows to re-use most of the existing configuration. 

To start the cli mode append `cli` as an command line argument, followed by the sub command and it's options.

example (docker):
```sh
$ docker run -i --rm ehrbase:next cli help

EHRbase CLI (Spring Boot 3.2.5 EHRbase  https://ehrbase.org/)
---------------------------------------------------------------------------
Help
---------------------------------------------------------------------------
Run with subcommand

cli [sub-command] [arguments]
s
Sub-commands:
    database  database related operation
    help      print this help message

Examples:

# show this help message
cli help

# show help message of a sub-command
cli database
cli database help

# execute sub-command with arguments
cli database --check-connection
```
 
The currently implementation support execution of flyway migrations by using

```sh
$ docker run -i --rm ehrbase:next cli database --migration-migrate
```

# Pre-Merge checklist

- [x] New code is tested
- [x] Changelog is updated
- [ ] Code has been reviewed

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,4 +11,4 @@ WORKDIR /app
 
 EXPOSE 8080
 
-ENTRYPOINT ["sh", "-c", "java ${JAVA_OPTS} -jar /app/ehrbase.jar --spring.profiles.active=docker"]
+ENTRYPOINT ["java", "-jar", "/app/ehrbase.jar", "--spring.profiles.active=docker"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,12 +3,17 @@ FROM eclipse-temurin:21-jre-alpine
 
 RUN addgroup -S ehrbase && adduser -S ehrbase -G ehrbase
 
-COPY /application/target/ehrbase.jar /app/ehrbase.jar
-
-RUN chown -R ehrbase:ehrbase /app
 USER ehrbase
+
 WORKDIR /app
+
+COPY /application/target/ehrbase.jar /app/ehrbase.jar
+COPY --chown=ehrbase:ehrbase /docker-entrypoint.sh /app/docker-entrypoint
+
+RUN chown -R ehrbase:ehrbase /app &\
+    chmod +x /app/docker-entrypoint
 
 EXPOSE 8080
 
-ENTRYPOINT ["java", "-jar", "/app/ehrbase.jar", "--spring.profiles.active=docker"]
+# wrapped in entrypoint to be able to accept cli args and use jacoco cli env var
+ENTRYPOINT ["/app/docker-entrypoint"]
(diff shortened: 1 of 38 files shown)


Dockerfile
```diff
@@ -11,4 +11,4 @@ WORKDIR /app
 
 EXPOSE 8080
 
-ENTRYPOINT ["sh", "-c", "java ${JAVA_OPTS} -jar /app/ehrbase.jar --spring.profiles.active=docker"]
```
vmueller-vg, line 14: Why?

application/src/main/java/org/ehrbase/application/server/EhrBaseServer.java
```diff
@@ -0,0 +1,46 @@
...
+            SecurityAutoConfiguration.class
+        })
+@Import({EhrBaseServerConfiguration.class})
+@SuppressWarnings("java:S1118")
+public class EhrBaseServer {
+
+    public static SpringApplication build(String[] args) {
+        return new SpringApplicationBuilder(EhrBaseServer.class)
+                .web(WebApplicationType.SERVLET)
+                .profiles("cli")
```
vmueller-vg, line 43: a "cli" profile for the server seems wrong

cli/src/main/java/org/ehrbase/cli/config/CliAqlQueryContext.java
```diff
@@ -0,0 +1,134 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+/**
+ * AQL query context implementation used for EHRbase CLI that can be configured using
+ * <code>-aql-mode=default|dry_run|show_executed_aql|show_executed_sql|show_query_plane</code> params.
+ */
+@Component(AqlQueryContext.BEAN_NAME)
+public class CliAqlQueryContext implements AqlQueryContext {
```
vmueller-vg, line 40: This should just throw `UnsupportedOperationException` on any method call, since executing AQL is not a feature of the CLI

vmueller-vg (changes requested):

application/src/main/java/org/ehrbase/application/server/EhrBaseServer.java
alexlehn: oops .. removed

Dockerfile
alexlehn: The problem was that this format did not accept arguments. But I forgot to support the `JAVA_OPTS` env var. I have extracted the command into an entrypoint script. This script now supports env vars and arguments.

cli/src/main/java/org/ehrbase/cli/config/CliAqlQueryContext.java
alexlehn: Was intended as a preparation in case we need cli AQL support. I have replaced the method bodies with `UnsupportedOperationException`.

vmueller-vg (approved):

alexlehn:
@vmueller-vg should we update the readme and doc as well?

PR #1390, opened 2024-08-14
CDR-1325 Enable Performance tests to Build and Test pipeline

vladislavploaia:


.github/workflows/build_and_test.yml
```diff
@@ -277,6 +278,179 @@ jobs:
...
+            docker-compose.yml
+
+      - name: Download - Test Image
+        uses: actions/download-artifact@v4
+        with:
+          name: ehrbase-image-test
+          path: ${{ runner.temp }}
+
+      - name: Docker - Load Image
+        run: docker load --input ${{ runner.temp }}/ehrbase-test.tar
```
alexlehn, line 315: I'm not entirely sure but possible we should upload and use the `ehrbase/ehrbase:build` image from here https://github.com/ehrbase/ehrbase/blob/feature/CDR-1325_perf_tests_pipeline/.github/workflows/build_and_test.yml#L120 ... the `ehrbase/ehrbase:test` image run with jacococo coverage agent that "could" have some negative performance impact.

.github/workflows/build_and_test.yml
```diff
@@ -422,6 +597,8 @@ jobs:
     needs: [
       coverage-collect,
       integration-test-collect,
+      performance-test-run,
```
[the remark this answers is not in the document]
vladislavploaia: fixed

.github/workflows/build_and_test.yml
vladislavploaia: fixed

alexlehn (approved):

PR #1449, opened 2025-01-03
CDR-1667 gracefully exit on future db version

alexlehn:
# Changes

Changed flyway config to gracefully fail in case an "older" ehrbase version is deployed against a newer database version.
In such cases EHRBase will now fail with the following message

```
Caused by: org.flywaydb.core.api.exception.FlywayValidateException: Validate failed: Migrations have failed validation
Detected applied migration not resolved locally: 20.
If you removed this migration intentionally, run repair to mark the migration as deleted.
Need more flexibility with validation rules? Learn more: https://rd.gt/3AbJUZE
```

An Alternative we could also update to Flyway 11 https://github.com/ehrbase/ehrbase/pull/1450


alexlehn:
@stefanspiska not sure if we should make this configurable and enabled by default, what do you think?

HolgerReiseVSys:
Was succeeded by #1450

PR #1450, opened 2025-01-03
CDR-1667 Update flyway to V11 to gracefully exit on future db version

alexlehn:
# Changes

Update to Flyway V11.

Changed flyway config to gracefully fail in case an "older" ehrbase version is deployed against a newer database version.
In such cases EHRBase will now fail with the following message

```
Caused by: org.flywaydb.core.api.exception.FlywayValidateException: Validate failed: Migrations have failed validation
Detected applied migration not resolved locally: 20.
If you removed this migration intentionally, run repair to mark the migration as deleted.
Need more flexibility with validation rules? Learn more: https://rd.gt/3AbJUZE
```

An Alternative solution for https://github.com/ehrbase/ehrbase/pull/1449

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -59,7 +59,7 @@
     <commons-io.version>2.18.0</commons-io.version>
     <commons-lang3.version>3.17.0</commons-lang3.version>
     <ehrbase.openehr.sdk.version>2.20.0</ehrbase.openehr.sdk.version>
-    <flyway.version>11.1.0</flyway.version>
+    <flyway.version>11.1.1</flyway.version>
     <jackson-bom.version>2.18.1</jackson-bom.version>
     <javamelody.version>1.99.1</javamelody.version>
     <jooq.version>3.19.15</jooq.version>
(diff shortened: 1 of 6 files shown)


bom/pom.xml
```diff
@@ -59,7 +59,7 @@
     <commons-io.version>2.18.0</commons-io.version>
     <commons-lang3.version>3.17.0</commons-lang3.version>
     <ehrbase.openehr.sdk.version>2.20.0</ehrbase.openehr.sdk.version>
-    <flyway.version>8.5.13</flyway.version>
+    <flyway.version>11.1.0</flyway.version>
```
HolgerReiseVSys, line 62: Version is outdated

HolgerReiseVSys (changes requested):

bom/pom.xml
alexlehn: or the PR ist old 😏

alexlehn:
@HolgerReiseVSys build errors are related to some issues on https://quay.io that returned 502 some times

HolgerReiseVSys (approved):

PR #1458, opened 2025-01-28
allow exception to cache proxying

stefanspiska:


service/src/main/java/org/ehrbase/cache/CacheProperties.java
```diff
@@ -26,6 +28,7 @@
 @ConfigurationProperties(prefix = "cache")
 public class CacheProperties {
 
+    private List<String> doNotProxy = new ArrayList<>();
```
vmueller-vg, line 31: a better name would be "txProxyExcludedBeanNames"

vmueller-vg (changes requested):

vmueller-vg (approved):

PR #1497, opened 2025-04-29
CDR-1901 - Remove EHRBase Template Id header

stephan-song:
# Changes

- Removed the non-standard template id header usages (usage not clear)
- Queries to retrieve this id may still happen as the id is used in auditing, for example

# Pre-Merge checklist

- [ ] New code is tested
- [ ] Changelog is updated
- [ ] Code has been reviewed

### DIFF
--- a/rest-openehr/src/main/java/org/ehrbase/rest/openehr/OpenehrCompositionController.java
+++ b/rest-openehr/src/main/java/org/ehrbase/rest/openehr/OpenehrCompositionController.java
@@ -435,8 +435,6 @@ public class OpenehrCompositionController extends BaseController implements Comp
             templateId = compositionService.retrieveTemplateId(compositionId);
         }
 
-        respHeaders.addIfAbsent(EHRbaseHeader.TEMPLATE_ID, templateId);
-
         HttpRestContext.register(
                 EHR_ID,
                 ehrId,
--- a/api/src/main/java/org/ehrbase/api/rest/EHRbaseHeader.java
+++ b/api/src/main/java/org/ehrbase/api/rest/EHRbaseHeader.java
@@ -24,8 +24,6 @@ public final class EHRbaseHeader {
 
     private EHRbaseHeader() {}
 
-    public static final String TEMPLATE_ID = "EHRBase-Template-ID";
-
     /**
      * Used by the /query endpoint to perform only a dry run query.
      */


rest-openehr/src/main/java/org/ehrbase/rest/openehr/OpenehrCompositionController.java
```diff
@@ -379,30 +384,25 @@ private <T extends CompositionResponseData> Optional<InternalResponse<T>> buildC
             CompositionRepresentation responseRepresentation,
             URI uri,
             List<String> headerList,
-            Supplier<T> factory) {
+            Supplier<T> factory,
+            Optional<String> availableTemplateId) {
```
stefanspiska, line 388: I'm not a fan of optional as parameter because you can not enforce that the optional is not null

stefanspiska: I would make @Nullable String templateIdFromContext and then do a null check later

stefanspiska (changes requested):

rest-openehr/src/main/java/org/ehrbase/rest/openehr/OpenehrCompositionController.java
```diff
@@ -127,15 +129,19 @@ public ResponseEntity createComposition(
...
                 responseRepresentation,
                 uri,
                 headerList,
-                RETURN_REPRESENTATION.equals(prefer) ? () -> new CompositionResponseData(null, null) : () -> null);
+                RETURN_REPRESENTATION.equals(prefer) ? () -> new CompositionResponseData(null, null) : () -> null,
+                Optional.ofNullable(compoObj)
+                        .map(Locatable::getArchetypeDetails)
+                        .map(Archetyped::getTemplateId)
+                        .map(ObjectId::getValue)
+                        .orElse(null));
```
HolgerReiseVSys, line 144: No need to use Optional: the template id is required

rest-openehr/src/main/java/org/ehrbase/rest/openehr/OpenehrCompositionController.java
```diff
@@ -217,8 +222,12 @@ public ResponseEntity updateComposition(
...
                     uri,
                     headerList,
-                    RETURN_REPRESENTATION.equals(prefer) ? () -> new CompositionResponseData(null, null) : () -> null);
-
+                    RETURN_REPRESENTATION.equals(prefer) ? () -> new CompositionResponseData(null, null) : () -> null,
+                    Optional.of(compoObj)
+                            .map(Locatable::getArchetypeDetails)
+                            .map(Archetyped::getTemplateId)
+                            .map(ObjectId::getValue)
+                            .orElse(null));
```
HolgerReiseVSys, line 230: No need to use Optional: the template id is required

rest-openehr/src/main/java/org/ehrbase/rest/openehr/OpenehrCompositionController.java
```diff
@@ -331,20 +340,18 @@ public ResponseEntity getComposition(
 
         URI uri = createLocationUri(EHR, ehrId.toString(), COMPOSITION, versionedObjectUid);
 
-        List<String> headerList = Arrays.asList(
-                LOCATION,
-                ETAG,
-                LAST_MODIFIED); // whatever is required by REST spec - CONTENT_TYPE only needed for 200, so handled
-        // separately
+        // whatever is required by REST spec - CONTENT_TYPE only needed for 200, so handled separately
+        var headerList = List.of(LOCATION, ETAG, LAST_MODIFIED);
```
HolgerReiseVSys, line 344: We usually only use var if generics make it extremely unreadable, as it is hard to review

rest-openehr/src/main/java/org/ehrbase/rest/openehr/OpenehrCompositionController.java
```diff
@@ -431,12 +434,10 @@ private <T extends CompositionResponseData> Optional<InternalResponse<T>> buildC
             respHeaders.setContentType(responseRepresentation.mediaType);
         } // else continue with returning but without additional data from above, e.g. body
 
-        if (isBlank(templateId)) {
-            templateId = compositionService.retrieveTemplateId(compositionId);
+        if (isBlank(templateId) && !isBlank(contextTemplateId)) {
+            templateId = contextTemplateId;
```
HolgerReiseVSys, line 438: if contextTemplateId is missing, we still need to ask the database

HolgerReiseVSys (changes requested):

rest-openehr/src/main/java/org/ehrbase/rest/openehr/OpenehrCompositionController.java
HolgerReiseVSys: Taking a second look, the template id seems to be always present.
`templateId ` could be set to `contextTemplateId` in the else-block of `if (minimalOrRepresentation != null)`.

rest-openehr/src/main/java/org/ehrbase/rest/openehr/OpenehrCompositionController.java
stephan-song: alright; I just used the Optional because my IDE suggested that the template id could produce a NPE (for both create and update)

rest-openehr/src/main/java/org/ehrbase/rest/openehr/OpenehrCompositionController.java
alexlehn: `templateId` is null when you come from `public ResponseEntity getComposition(...)`

rest-openehr/src/main/java/org/ehrbase/rest/openehr/OpenehrCompositionController.java
```diff
@@ -379,7 +381,8 @@ private <T extends CompositionResponseData> Optional<InternalResponse<T>> buildC
             CompositionRepresentation responseRepresentation,
             URI uri,
             List<String> headerList,
-            Supplier<T> factory) {
+            Supplier<T> factory,
```
alexlehn, line 384: @stephan-song because you need to rebase anyway - could you replace this `Supplier<T> factory` parameter with the value of `T` - I did not see a reason to have a factory supplier instead of a simple argument - seems to be a leftover from an older refactoring.

PR #1524, opened 2025-08-05
Contains expression semantics

razvan-grq:
# Changes

> Improve CONTAINS expression semantics when using at-code predicates.
If the parent CONTAINS expression is an archetype, it must be located within that archetype.
If the parent CONTAINS expression is an at-code, it must be located within the same parent archetype and be a descendant.

# Pre-Merge checklist

- [ ] New code is tested
- [ ] Changelog is updated
- [ ] Code has been reviewed

### DIFF
--- a/aql-engine/src/main/java/org/ehrbase/openehr/aqlengine/querywrapper/AqlQueryWrapper.java
+++ b/aql-engine/src/main/java/org/ehrbase/openehr/aqlengine/querywrapper/AqlQueryWrapper.java
@@ -170,10 +170,22 @@ public final class AqlQueryWrapper {
         final List<ContainsWrapper> chain = new ArrayList<>();
         final ContainsSetOperationWrapper setOperator;
 
+        ContainsWrapper parent = null;
         Containment next = root;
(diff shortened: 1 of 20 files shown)


aql-engine/src/main/java/org/ehrbase/openehr/aqlengine/querywrapper/AqlQueryWrapper.java
```diff
@@ -170,10 +170,22 @@ private static ContainsChain buildContainsChain(
...
 
+        ContainsWrapper parent = null;
         Containment next = root;
         while (next instanceof AbstractContainmentExpression c) {
 
-            chain.add(containsDescs.get(next));
+            ContainsWrapper current = containsDescs.get(next);
+            chain.add(current);
+
+            // TODO: could have a setter on the contains wrapper?
```
HolgerReiseVSys, line 180: Just do it :D

HolgerReiseVSys (changes requested):
We need to mention the change in the changelog.
It would make sense to add ASL graph tests for multiple scenarions (incl. nested and/or)

